PR #1, opened 2021-04-09
Moved logger from dapr/dapr

pkedy:
Direct copy from dapr/dapr/pkg/logger with the exception of removing the dependency on dapr/dapr/pkg/version. I replaced that with a variable that must be set called `DaprVersion`.


pkedy:
Forgot the Makefile and github actions workflow.  Adding now.

Makefile
```diff
@@ -0,0 +1,85 @@
...
+
+################################################################################
+# Target: check-diff                                                           #
+################################################################################
+.PHONY: check-diff
+check-diff:
+	git diff --exit-code ./go.mod # check no changes
+
+################################################################################
+# Target: conf-tests                                                           #
```
artursouza, line 81: Remove this, please.

.github/workflows/kit.yml
```diff
@@ -0,0 +1,67 @@
...
+# Copyright (c) Microsoft Corporation and Dapr Contributors.
+# Licensed under the MIT License.
+# ------------------------------------------------------------
+
+name: kit
+
+on:
+  push:
+    branches:
+      - master
```
artursouza, line 11: `master` branch is not called `main` - this probably explains why PR action did not get triggered.

logger/dapr_logger.go
```diff
@@ -0,0 +1,146 @@
...
+		}),
+	}
+
+	dl.EnableJSONOutput(defaultJSONOutput)
+
+	return dl
+}
+
+// EnableJSONOutput enables JSON formatted output log
+func (l *daprLogger) EnableJSONOutput(enabled bool) {
```
artursouza, line 40: Please, get the version as param instead.

logger/dapr_logger.go
```diff
@@ -0,0 +1,146 @@
...
+
+// daprLogger is the implemention for logrus
+type daprLogger struct {
+	// name is the name of logger that is published to log as a scope
+	name string
+	// loger is the instance of logrus logger
+	logger *logrus.Entry
+}
+
+var GetVersion = func() string {
```
artursouza, line 18: remove this global var and get version via param in the method where this is used. See comment below.

artursouza (changes requested):

logger/dapr_logger.go
pkedy: The problem with is that is called by `newDaprLogger` which is not exposed by the package. I propose keeping this as a global variable (change from func to a string variable) to keep the changes in dapr/dapr to a minimum. There is a much more elegant way of handling this but it is outside the scope of this PR.

artursouza: I see it. OK, we can refactor that later.

artursouza (approved):

Issue #2, opened 2021-04-29
Add kit to CodeCov report


artursouza:
## Describe the proposal

Report on unit test coverage using CodeCov and make it available in the report below:

https://codecov.io/gh/dapr

PR #4, opened 2021-06-02
ci: add codecov

daixiang0:
Signed-off-by: Long Dai <long0dai@foxmail.com>

# Description

Add codecov.

## Issue reference

Please reference the issue this PR will close: #2

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/.codecov.yaml
+++ b/.codecov.yaml
@@ -1,0 +1,11 @@
+coverage:
+  # Commit status https://docs.codecov.io/docs/commit-status are used
+  # to block PR based on coverage threshold.
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 0%
+    patch:
+      default:
+        informational: true


.codecov.yaml
```diff
@@ -0,0 +1,11 @@
+coverage:
+  # Commit status https://docs.codecov.io/docs/commit-status are used
+  # to block PR based on coverage threshold.
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 0%
```
pkedy, line 8: Since this is a new repo, we can start out with "ideal" thresholds.  I'm wondering if 0% is too aggressive
for Golang. If you add more error cases to a file the coverage can drop slightly (which I think is OK). But IMO, the overall coverage should never drop below 80%. In my exp, this has been a pretty good policy for Go. Thoughts?

pkedy (commented):
Hi @daixiang0!  Thanks for this PR.  We definitely need codecov here.  What do you think about my comment below?

.codecov.yaml
artursouza: LGTM. We can always use admin override when this is violated.

artursouza (approved):

.codecov.yaml
pkedy: And we can always adjust the policy later.

pkedy (approved):

.codecov.yaml
daixiang0: Still need update thresholds? I can do it. @artursouza @pkedy

PR #9, opened 2021-10-13
feature/log: add traceid and file for log 

1046102779:
https://github.com/dapr/dapr/issues/3759


1046102779:
@pkedy please help me review this PR. Thanks.

1046102779:
In order to troubleshoot problems more easily, And we need to output traceid and other infos in log.

Logger interface need to  add two methods included  WithTrac and WithContext.

At the same time, these are compatible with different versions of systems and logic servers.

1046102779:
@artursouza @yaron2 @pkedy +1

1046102779:
Another feature:
Added log file output scheme for daprd runtime, adding two fields: dapr.io/log-output and dapr.io/log-path, which respectively indicate the log type (console and file, default console). When the log type is file, log-path is enabled, indicating the log file path.

1046102779:
@artursouza @yaron2 @pkedy. please help me review this PR. The PR has been around for a long time, it’s more important to me

yaron2:
> @artursouza @yaron2 @pkedy. please help me review this PR. The PR has been around for a long time, it’s more important to me

We will look at it this week. Sorry for the long delay.

yaron2:
@pkedy can you please review this?

logger/dapr_logger.go
```diff
@@ -1,21 +1,32 @@
...
 // daprLogger is the implemention for logrus.
 type daprLogger struct {
 	// name is the name of logger that is published to log as a scope
 	name string
 	// loger is the instance of logrus logger
-	logger *logrus.Entry
+	logger  *logrus.Entry
+	ctx     context.Context
+	traceid string
+	sync.Mutex
```
daixiang0, line 23: Do we need it here?

logger/logger.go
```diff
@@ -25,6 +26,7 @@ const (
 	logFieldInstance  = "instance"
 	logFieldDaprVer   = "ver"
 	logFieldAppID     = "app_id"
+	logFieldTrace     = "id"
```
daixiang0, line 29: Better named as `trace_id`

logger/dapr_logger.go
```diff
@@ -93,52 +104,110 @@ func (l *daprLogger) WithLogType(logType string) Logger {
...
+func (l *daprLogger) WithContext(ctx context.Context) Logger {
+	return &daprLogger{
+		name:   l.name,
+		logger: l.logger,
+		ctx:    ctx,
+	}
+}
+
+// withTraceFromContext add trace id to log by context.
+func (l *daprLogger) withTraceFromContext() {
```
daixiang0, line 136: Is it just a helper function?

logger/dapr_logger.go
```diff
@@ -93,52 +104,110 @@ func (l *daprLogger) WithLogType(logType string) Logger {
...
+	if !trace.IsValid(sc.TraceID) {
+		return
+	}
+	l.logger = l.logger.WithField(logFieldTrace, sc.TraceID.String())
+	l.traceid = sc.TraceID.String()
+}
+
 // Info logs a message at level Info.
 func (l *daprLogger) Info(args ...interface{}) {
+	l.withTraceFromContext()
```
daixiang0, line 157: One call would call one lock, looks inefficient.

1046102779:
```golang
l.logger = l.logger.WithField(logFieldTrace, sc.TraceID.String())
```

Lock is used because WithFields has concurrency safety issues. Is there any other good way?

企业微信截图_16357682451239

pkedy:
Hi @1046102779, sorry for the delay. I want to make sure I understand the approach you are taking.

Indeed, logging the traceID (and/or other IDs for the request) would be very beneficial. The way Dapr currently handles logging is via global loggers. They are not passed on the stack. Based on that I have a concern in `withTraceFromContext` because it is changing the underlying Logrus logger variable. Since the logger is global, I believe that posses a problem. Also, the lock does not encapsulate the actual log action so I think there is a race condition. Lastly, if Logrus is creating a stack of loggers, then this would consume/leak memory. Did I misunderstand the approach?

This potentially gets into the topic of how the Dapr loggers _should_ be passed. Here is a great article that outlines the possible approaches.

https://gogoapps.io/blog/passing-loggers-in-go-golang-logging-best-practices/

I'm my previous projects, I have passed the logger via context, adding request level fields as the logger flows through the call stack. Some might be consider this poor design because the dependency on the logger is not explicit but I have not found a more elegant way to solve this.

If we use context to pass the logger, we would need to have a `logger.WithField(key, value)` method that returns a new `daprLogger/Logger` and a pair of `WithContext`/`FromContext` methods to store and extract the logger from context. This also would involve some adjustments to the Dapr API code to pull the logger from context instead of the global variable. This might be pretty simple to do but it would still require careful thought. @yaron2 

So stepping back: Did I misunderstand the design of this PR? Do you have any thoughts on the above or perhaps a more elegant solution?

1046102779:
> Lastly, if Logrus is creating a stack of loggers, then this would consume/leak memory. Did I misunderstand the approach?

Q: `Lastly, if Logrus is creating a stack of loggers, then this would consume/leak memory. Did I misunderstand the approach?`
A: It won't consume/leak memory. In logrus, it wll override key field. 

logurs/entry.go:125
```golang
125 func (entry *Entry) WithFields(fields Fields) *Entry {
126         data := make(Fields, len(entry.Data)+len(fields))
127         for k, v := range entry.Data {
128                 data[k] = v
129         }
130         fieldErr := entry.err
131         for k, v := range fields {
```

Q: `So stepping back: Did I misunderstand the design of this PR? Do you have any thoughts on the above or perhaps a more elegant solution?`
 A: You have no misunderstanding, I want to do two things: one is to add trace and context; the other is that kit not only supports console but also file reporting. 

Q: `https://gogoapps.io/blog/passing-loggers-in-go-golang-logging-best-practices/`
A:  Thanks. I will read this article first to see if there is any better solution, but I think the current solution should be feasible.

pkedy:
After looking at this a bit more, the usage makes sense. I’m still hung up on if the lock is needed and a few other things. I will look at this again tomorrow but since we are likely in different timezones, I wanted to let you know.

1046102779:
> After looking at this a bit more, the usage makes sense. I’m still hung up on if the lock is needed and a few other things. I will look at this again tomorrow but since we are likely in different timezones, I wanted to let you know that this has my attention.

Thanks. There may be some problems with logrus entry's WithFields. I feel that it should check whether to expand before expanding. Because there may be some identical keys, only the value needs to be overwritten.


```golang
125 func (entry *Entry) WithFields(fields Fields) *Entry {
126         data := make(Fields, len(entry.Data)+len(fields))
127         for k, v := range entry.Data {
128                 data[k] = v
129         }
```

After changed.

```golang
124 // Add a map of fields to the Entry.
  125 func (entry *Entry) WithFields(fields Fields) *Entry {
~ 126         var count int = len(entry.Data)
+ 127         for k := range fields {
+ 128                 if _, ok := entry.Data[k]; !ok {
+ 129                         count++
+ 130                 }
+ 131         }
+ 132         data := make(Fields, count)
  133         for k, v := range entry.Data {
  134                 data[k] = v
  135         }
```

1046102779:
> After looking at this a bit more, the usage makes sense. I’m still hung up on if the lock is needed and a few other things. I will look at this again tomorrow but since we are likely in different timezones, I wanted to let you know that this has my attention.
> 
> P.S. I don’t think there is a concurrency issue in entry.go of Logrus because entry.Data probably is not mutated after it is copied into a new Entry. Thoughts?

Today i verify whether there is a concurrency problem with the logrus entry. If there is no problem, I will revoke the lock code

daixiang0:
Suddenly think that could we use sync map to store those KV info as we want read lock in most cases?

1046102779:
> Suddenly think that could we use sync map to store those KV info as we want read lock in most cases?

It means to use the global logger to store all Fields, and then when writing logs, and attach these fields?

daixiang0:
Just an idea if we want read lock indeed.

1046102779:
> Hi @1046102779, sorry for the delay. I want to make sure I understand the approach in this PR.
> 
> Indeed, logging the traceID (and/or other IDs for the request) would be very beneficial. The way Dapr currently handles logging is via global loggers. They are not passed on the stack. Based on that I have a concern in `withTraceFromContext` because it changes the underlying Logrus logger variable. Since the logger is global, I believe that posses a problem. Also, the lock does not encapsulate the actual log action so I think there is a race condition. Lastly, if Logrus is creating a stack of loggers, then this would consume/leak memory. Did I misunderstand the approach?
> 
> This potentially gets into the topic of how the Dapr loggers _should_ be passed. Here is a great article that outlines the possible approaches.
> 
> https://gogoapps.io/blog/passing-loggers-in-go-golang-logging-best-practices/
> 
> I'm my previous projects, I have passed the logger via context, adding request level fields as the logger flows through the call stack. Some might be consider this poor design because the dependency on the logger is not explicit but I have not found a more elegant way to solve this.
> 
> If we use context to pass the logger, we would need to have a `logger.WithField(key, value)` method that returns a new `daprLogger/Logger` and a pair of `WithContext`/`FromContext` methods to store and extract the logger from context. This also would involve some adjustments to the Dapr API code to pull the logger from context instead of the global variable. This might be pretty simple to do but it would still require careful thought. @yaron2
> 
> So stepping back: Did I misunderstand the design of this PR? Do you have any thoughts on the above or perhaps a more elegant solution?

Q: `https://gogoapps.io/blog/passing-loggers-in-go-golang-logging-best-practices/`
A:  The two are not the same. Traceid naturally exists in the context, while logger generally does not exist in the context. The two are not the same.

pkedy:
Hello again, @1046102779! Based on your usage examples (summarized below) I propose simplifying this a little.

```golang
log := apiServerLogger.WithContext(ctx)
// or
log:= apiServerLogger.WithTrace(traceid)
log.Info(err)
```

I'm very confident the lock is unnecessary and thus `withTraceFromContext` is not needed. If `WithContext` were implemented like this (just tweaking your code slightly), I believe it satisfies the the above scenarios.

```golang
// WithContext write context infos to log.
func (l *daprLogger) WithContext(ctx context.Context) Logger {
	sc := trace.GetSpanContext(l.ctx)
	// check trace invalid
	if trace.IsValid(sc.TraceID) {
		return &daprLogger{
			name:   l.name,
			logger: l.logger.WithField(logFieldTrace, sc.TraceID.String()),
			ctx:    ctx,
		}
	}

	// Return the unchanged logger
	return l
}
```

That said, I don't like mixing logging and tracing concerns together in the same component. To maintain the [single-responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle), I recommend a "helper" function somewhere in dapr/dapr. Something like this.

```golang
const logFieldTraceID = "trace_id"

// LoggerWithContext returns a new logger that contains fields for information from the context.
func LoggerWithContext(ctx context.Context, log logger.Logger) logger.Logger {
	sc := trace.GetSpanContext(ctx)
	// check trace invalid
	if trace.IsValid(sc.TraceID) {
		return log.WithField(logFieldTraceID, sc.TraceID.String())
	}

	// Return the unchanged logger
	return log
}
```

And add `WithField/WithFields` to the dapr logger.

```golang
func (l *daprLogger) WithField(key string, value interface{}) Logger {
	return &daprLogger{
		name:   l.name,
		logger: l.logger.WithField(key, value),
	}
}

func (l *daprLogger) WithFields(fields map[string]interface{}) Logger {
	return &daprLogger{
		name:   l.name,
		logger: l.logger.WithFields(fields),
	}
}
```

Another benefit is that we no logger need to put the tracing & gRPC dependencies in dapr/kit for this purpose.

1046102779:
> > After looking at this a bit more, the usage makes sense. I’m still hung up on if the lock is needed and a few other things. I will look at this again tomorrow but since we are likely in different timezones, I wanted to let you know that this has my attention.
> 
> Thanks. There may be some problems with logrus entry's WithFields. I feel that it should check whether to expand before expanding. Because there may be some identical keys, only the value needs to be overwritten.
> 
> ```go
> 125 func (entry *Entry) WithFields(fields Fields) *Entry {
> 126         data := make(Fields, len(entry.Data)+len(fields))
> 127         for k, v := range entry.Data {
> 128                 data[k] = v
> 129         }
> ```
> 
> After changed.
> 
> ```go
> 124 // Add a map of fields to the Entry.
>   125 func (entry *Entry) WithFields(fields Fields) *Entry {
> ~ 126         var count int = len(entry.Data)
> + 127         for k := range fields {
> + 128                 if _, ok := entry.Data[k]; !ok {
> + 129                         count++
> + 130                 }
> + 131         }
> + 132         data := make(Fields, count)
>   133         for k, v := range entry.Data {
>   134                 data[k] = v
>   135         }
> ```


> > After looking at this a bit more, the usage makes sense. I’m still hung up on if the lock is needed and a few other things. I will look at this again tomorrow but since we are likely in different timezones, I wanted to let you know that this has my attention.
> 
> Thanks. There may be some problems with logrus entry's WithFields. I feel that it should check whether to expand before expanding. Because there may be some identical keys, only the value needs to be overwritten.
> 
> ```go
> 125 func (entry *Entry) WithFields(fields Fields) *Entry {
> 126         data := make(Fields, len(entry.Data)+len(fields))
> 127         for k, v := range entry.Data {
> 128                 data[k] = v
> 129         }
> ```
> 
> After changed.
> 
> ```go
> 124 // Add a map of fields to the Entry.
>   125 func (entry *Entry) WithFields(fields Fields) *Entry {
> ~ 126         var count int = len(entry.Data)
> + 127         for k := range fields {
> + 128                 if _, ok := entry.Data[k]; !ok {
> + 129                         count++
> + 130                 }
> + 131         }
> + 132         data := make(Fields, count)
>   133         for k, v := range entry.Data {
>   134                 data[k] = v
>   135         }
> ```

@pkedy  Using withfields method will expand rapidly slice? Help me check the point.

pkedy:
If your concern is this line:

```golang
data := make(Fields, len(entry.Data)+len(fields))
```

I would not be concerned because `Fields` is a map, not a slice. The size argument for maps is merely a hint as to how many underlying hash buckets to create. The code first is making a copy of the existing fields map, then overlays the fields passed in which may override a previous field value. More often than not, each of the added values are unique, so `len(entry.Data)+len(fields)` is a good calculation to use. Again, because it is a map, it only needs to be an approximation.

https://qvault.io/golang/golang-make-maps-and-slices/

Do this help?

1046102779:
> copy of the existing fields map, then overlays the fields passed in which may override a previous field value. More often than not, each of the added values are unique, so `len(entry.Data)+len(fields)` is a good calculation to use. Again, becaus

Yes. I verify that there is no problem.

wecom-temp-ca9ae2a4d002c6ac33b11053ce539251
wecom-temp-51906846a354cd68c9c62377157e0826

1046102779:
> Hello again, @1046102779! Based on your usage examples (summarized below) I propose simplifying this a little.
> 
> ```go
> log := apiServerLogger.WithContext(ctx)
> // or
> log:= apiServerLogger.WithTrace(traceid)
> log.Info(err)
> ```
> 
> I'm very confident the lock is unnecessary and thus `withTraceFromContext` is not needed. If `WithContext` were implemented like this (just tweaking your code slightly), I believe it satisfies the the above scenarios.
> 
> ```go
> // WithContext write context infos to log.
> func (l *daprLogger) WithContext(ctx context.Context) Logger {
> 	sc := trace.GetSpanContext(l.ctx)
> 	// check trace invalid
> 	if trace.IsValid(sc.TraceID) {
> 		return &daprLogger{
> 			name:   l.name,
> 			logger: l.logger.WithField(logFieldTrace, sc.TraceID.String()),
> 			ctx:    ctx,
> 		}
> 	}
> 
> 	// Return the unchanged logger
> 	return l
> }
> ```
> 
> That said, I don't like mixing logging and tracing concerns together in the same component. To maintain the [single-responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle), I recommend a "helper" function somewhere in dapr/dapr. Something like this.
> 
> ```go
> const logFieldTraceID = "trace_id"
> 
> // LoggerWithContext returns a new logger that contains fields for information from the context.
> func LoggerWithContext(ctx context.Context, log logger.Logger) logger.Logger {
> 	sc := trace.GetSpanContext(ctx)
> 	// check trace invalid
> 	if trace.IsValid(sc.TraceID) {
> 		return log.WithField(logFieldTraceID, sc.TraceID.String())
> 	}
> 
> 	// Return the unchanged logger
> 	return log
> }
> ```
> 
> And add `WithField/WithFields` to the dapr logger.
> 
> ```go
> func (l *daprLogger) WithField(key string, value interface{}) Logger {
> 	return &daprLogger{
> 		name:   l.name,
> 		logger: l.logger.WithField(key, value),
> 	}
> }
> 
> func (l *daprLogger) WithFields(fields map[string]interface{}) Logger {
> 	return &daprLogger{
> 		name:   l.name,
> 		logger: l.logger.WithFields(fields),
> 	}
> }
> ```
> 
> Another benefit is that we no logger need to put the tracing & gRPC dependencies in dapr/kit for this purpose.

@pkedy  Two parts:

1. dapr/kit
```golang
type Logger interface {
 // add three methods
WithContext(ctx context.Context) Logger
WithField(key string, value interface{}) Logger
WithFields(fields map[string]interface{}) Logger
}
```

2. dapr/dapr

```golang
//  LoggerWithContext is a helper function.
LoggerWithContext(ctx context.Context, log logger.Logger) logger.Logger
```

Do we still need the WithTrace method?

1046102779:
@pkedy The code has been submitted.

logger/dapr_logger.go
```diff
@@ -4,6 +4,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/natefinch/lumberjack"
```
pkedy, line 7: Should this be `gopkg.in/natefinch/lumberjack.v2`? That might clean up `go.mod` a bit.

Something to note: This package was last maintained in 2017.

pkedy (commented):
This looks good, @1046102779! There are a few comments where @yaron2 and @artursouza should probably chime in.

yaron2:
> /cc @artursouza @yaron2

Taking a look

1046102779:
@pkedy  @artursouza @yaron2 Can we confirm this plan as soon as possible and use this feature in the v1.6 version? It is a important feature for developers

pkedy:
@1046102779 I can at least get the dapr/kit part of this merged. You would need to narrow the new Logger methods to `WithField`/`WithFields ` and remove the tracing logic and in favor of moving that to dapr/dapr.

1046102779:
> @1046102779 I can at least get the dapr/kit part of this merged. You would need to narrow the new Logger methods to `WithField`/`WithFields ` and remove the tracing logic and in favor of moving that to dapr/dapr.

Thank you so much pkedy.

What do you think of the way that Context obtains traces and adds and reports logs to files?

I think the collection of service logs and daprd runtime logs is also a very important feature in the log retrieval platform. When pod is destroyed, daprd runime trace is lost. This log needs to be persisted, which is convenient to locate the daprd runtime ->building Block->external service trace in time-consuming situation, and time-consuming analysis inside daprd runtime

1046102779:
> @1046102779 I can at least get the dapr/kit part of this merged. You would need to narrow the new Logger methods to `WithField`/`WithFields ` and remove the tracing logic and in favor of moving that to dapr/dapr.

At the same time, the new log file persistence method is not the default option. Users can use the default console or file persistence according to their needs

pkedy:
I could be mistaken how the cloud providers are gathering logs but in the case of my previous company running Kubernetes, Fluentd was used to collect container logs written to **stdout** and append them into Prometheus/etc. I've always thought this was the cloud native way of doing it and that stateless containers never write to disk.

That said, I'm OK having it as an option at least at the scope of this package. You'll need to get @yaron2 and @artursouza 's opinion on having file logs be an option in daprd.

1046102779:
> I could be mistaken how the cloud providers are gathering logs but in the case of my previous company running Kubernetes, Fluentd was used to collect container logs written to **stdout** and append them into Prometheus/etc. I've always thought this was the cloud native way of doing it and that stateless containers never write to disk.
> 
> That said, I'm OK having it as an option at least at the scope of this package. You'll need to get @yaron2 and @artursouza 's opinion on having file logs be an option in daprd.

Currently we are using this architecture.

image

artursouza:
I am OK to have logs to optionally be forwarded to logs. On the other hand, it would be interesting to also have built-in log rotation. So, we can have hourly or daily log files - which should also be configurable.

PR #18, opened 2022-07-26
Added to package to convert to pointers

ItalyPaleAle:
# Description

This added a new package to dapr/kit called `ptr` which contains a method to convert any value to a pointer, based on generics. (The code for that function is copied from the Azure SDK, because of the Go's philosophy ["a little copying is better than a little dependency"](https://go-proverbs.github.io/) - the MIT license permits that freely)

Currently Dapr was using agrea/ptr for that purpose, which didn't use generics so it was limited to the types that were explicitly supported. Since we use Go 1.18 now and can rely on generics, the new approach is more convenient.

Also, this PR updates dependencies and the linter.

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


pkedy:
@ItalyPaleAle I am supportive of this, but I have a question before merging. There are only a [handful](https://github.com/dapr/dapr/search?q=agrea%2Fptr) of replacements that could be made in dapr/dapr. Where else could this be beneficial?

pkedy (approved):
@ItalyPaleAle I am supportive of this, but I have a question before merging. There are only a [handful](https://github.com/dapr/dapr/search?q=agrea%2Fptr) of replacements that could be made in dapr/dapr. Where else could this be beneficial?

ItalyPaleAle:
@pkedy  this is one example in a PR where I needed to use the method from the Azure SDK because I needed a pointer to a time.Duration: https://github.com/dapr/dapr/pull/4903/files#diff-64faf11a53da287dad97fd63381370362fea01d0d9f9f3a8f4812c4a52b873c2R101 

Not a critical thing, of course. During development, I needed to change those values without changing the other tests (then reverted so the values are the same, so technically I could just revert the change and this wouldn't be needed at all - but it could still be useful).

PR #19, opened 2022-08-12
Enable custom output destination in Dapr Kit Logger

berndverst:
Signed-off-by: Bernd Verst <4535280+berndverst@users.noreply.github.com>

# Description

This allows a custom output destination in Dapr Kit logger which is very handy for writing tests, allowing the use of for example a byte buffer instead of stdout.

Essentially this exposes the `logrus` method `SetOutput(output io.Writer)`

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -68,7 +68,7 @@ type Logger interface {
 	EnableJSONOutput(enabled bool)
 
 	// Writes log messages to the specified io.Writer instead of stdout.
-	SetOuput(output io.Writer)
+	SetOutput(output io.Writer)
 
 	// SetAppID sets dapr_id field in the log. Default value is empty string
 	SetAppID(id string)
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -14,6 +14,7 @@ limitations under the License.
 package logger
 
 import (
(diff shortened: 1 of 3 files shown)


logger/logger.go
```diff
@@ -66,6 +67,9 @@ type Logger interface {
 	// EnableJSONOutput enables JSON formatted output log
 	EnableJSONOutput(enabled bool)
 
+	// Writes log messages to the specified io.Writer instead of stdout.
+	SetOuput(output io.Writer)
```
shubham1172, line 71:
```suggestion
	SetOutput(output io.Writer)
```

shubham1172: Or should we call it `SetWriter`?

berndverst:
@shubham1172 I fixed by typo. I want to keep the name as is because this is what the popular `logrus` utility which we are wrapping calls this. It is more consistent with what folks would expect.

mukundansundar (approved):

berndverst:
Since nobody looked at this for 2 months and this now has conflicts I am closing this PR.

PR #20, opened 2022-09-06
feature(trace): add context to Logger API

1046102779:
Signed-off-by: 1046102779 <seachen@tencent.com>

# Description

At present, our servers need to correlate logs and traces.

This is a very urgent matter, I hope everyone can review these code at quickly. Thanks.

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [ ] Logger API with context.
* [ ] Support Opentelemetry.
* [ ] Unit tests.

/cc @pkedy @yaron2 @artursouza @daixiang0 @berndverst @ItalyPaleAle

PR #21, opened 2022-09-14
feature: add context log

1046102779:
Signed-off-by: 1046102779 <seachen@tencent.com>

```golang
// NewContext returns a new Context, derived from ctx, which carries the provided Logger.
func NewContext(ctx context.Context, logger Logger) context.Context {
	return context.WithValue(ctx, contextKey{}, logger)
}

// FromContextOrDiscard returns a Logger from ctx.  If no Logger is found, this
// returns a Logger that discards all log messages.
func FromContextOrDefault(ctx context.Context) Logger {
	if v, ok := ctx.Value(contextKey{}).(Logger); ok {
		return v
	}

	return emptyOpLogger
}
```

/cc @pkedy   @daixiang0 @yaron2 @artursouza


ItalyPaleAle:
@1046102779 Can you clarify what you need this for?

1046102779:
> @1046102779 Can you clarify what you need this for?

When a microservice receives a request, not every function called has a logger object. In general, a global logger variable is defined in each `pkg` or `cmd` subdirectory of the `dapr` repo; the `contribsLog` variable is passed to each component in `components-contrib` during the initialization phase;

However, we may also encounter this situation:

When a method called does not have a log object, it will mix the business parameter list with the log parameter in the input parameters, which is very strange. It is recommended to use the context to pass the upstream log object, so that when a request comes, the context can be used. By concatenating the entire timeout control mechanism, the entire request chain can also use the same log object to generate the log stream. The purpose is to use the context to get all the information that has nothing to do with the business itself.

From another point of view, we should define as few global variables as possible. All variables should be elements of an object, and others are implemented by passing parameters. It is reasonable to put non-business-related parameters in the context. , even in the initialization phase, we can complete it through the context's timeout control and carry the log object. rather than through coarse-grained channels.

```shell
./runtime/pubsub/subscriptions.go:164:func filterSubscriptions(subscriptions []Subscription, log logger.
Logger) []Subscription {                                                                              
./runtime/pubsub/subscriptions.go:253:func DeclarativeSelfHosted(componentsPath string, log logger.Logge
r) (subs []Subscription) {                                                                              
./runtime/pubsub/subscriptions.go:432:func DeclarativeKubernetes(client operatorv1pb.OperatorClient, pod
Name string, namespace string, log logger.Logger) []Subscription {
./pkg/resiliency/resiliency.go:195:func LoadStandaloneResiliency(log logger.Logger, runtimeID, path string) []*resiliencyV1alpha.Resiliency {
./pkg/resiliency/resiliency.go:246:func LoadKubernetesResiliency(log logger.Logger, runtimeID, namespace string, operatorClient operatorv1pb.OperatorClient) []*resiliencyV1alpha.Resiliency {
./pkg/resiliency/resiliency.go:276:func FromConfigurations(log logger.Logger, c ...*resiliencyV1alpha.Resiliency) *Resiliency {
./pkg/resiliency/resiliency.go:295:func New(log logger.Logger) *Resiliency {
./pkg/resiliency/resiliency.go:952:func (e *circuitBreakerInstances) Get(log logger.Logger, instanceName string, template *breaker.CircuitBreaker) *breaker.CircuitBreaker {
./pkg/resiliency/policy.go:37:func Policy(ctx context.Context, log logger.Logger, operationName string, t time.Duration, r *retry.Config, cb *breaker.CircuitBreaker) Runner {
./pkg/components/state/pluggable.go:444:func fromConnector(_ logger.Logger, connector *pluggable.GRPCCon
nector[stateStoreClient]) *grpcStateStore {
./pkg/components/state/pluggable.go:452:func fromPluggable(l logger.Logger, pc components.Pluggable) *gr
pcStateStore {
./pkg/components/state/pluggable.go:457:func NewGRPCStateStore(l logger.Logger, socketFactory func(strin
g) string) *grpcStateStore {
./pkg/components/state/pluggable.go:462:func newGRPCStateStore(l logger.Logger, pc components.Pluggable)
 state.Store {
./pkg/components/pubsub/pluggable.go:190:func fromConnector(l logger.Logger, connector *pluggable.GRPCCo
nnector[proto.PubSubClient]) *grpcPubSub {
./pkg/components/pubsub/pluggable.go:199:func NewGRPCPubSub(l logger.Logger, socketFactory func(string)
string) *grpcPubSub {
./pkg/components/pubsub/pluggable.go:204:func newGRPCPubSub(l logger.Logger, pc components.Pluggable) pu
bsub.PubSub {
./pkg/components/pluggable/registry.go:28:func AddRegistryFor[T any](cmpType components.PluggableType, r
egFunc func(componentFactory func(logger.Logger) T, names ...string), factory func(logger.Logger, compon
ents.Pluggable) T)
```

```shell
./state/cockroachdb/cockroachdb_query.go:136:func (q *Query) execute(logger logger.Logger, db *sql.DB) (
[]state.QueryItem, string, error) {
./internal/authentication/azure/storage.go:39:func GetAzureStorageBlobCredentials(log logger.Logger, acc
ountName string, metadata map[string]string) (azblob.Credential, *azure.Environment, error) {
./internal/authentication/azure/storage.go:83:func GetAzureStorageQueueCredentials(log logger.Logger, ac
countName string, metadata map[string]string) (azqueue.Credential, *azure.Environment, error) {
```

yaron2:
I think passing a logger from context is a valid design. cc @art

1046102779:
> 

cc @artursouza

yaron2:
@daixiang0 can you help review?

1046102779:
> @daixiang0 can you help review?

@daixiang0 And this [PR#20](https://github.com/dapr/kit/pull/20) :)

logger/logger_test.go
```diff
@@ -74,3 +75,27 @@ func TestToLogLevel(t *testing.T) {
...
+	})
+
+	t.Run("input non-nil logger", func(t *testing.T) {
+		testLoggerName := "dapr.test"
+		logger := NewLogger(testLoggerName)
+		assert.NotNil(t, logger)
+
+		ctx := NewContext(context.Background(), logger)
+		assert.NotNil(t, ctx, "ctx is not nil")
+		logger02 := FromContextOrDefault(ctx)
```
daixiang0, line 96: `logger2`

logger/empty_logger.go
```diff
@@ -0,0 +1,47 @@
+package logger
+
+type emptyLogger struct{}
```
daixiang0, line 3: If only use in this package, keep initials lowercase is enough.

Better to name it as `default` and use `defalutlogger` in other places.

1046102779:
cc @daixiang0 . Help me check this PR. Thanks.

logger/empty_logger.go
daixiang0: ?

1046102779:
cc @daixiang0.

Help me review again.

yaron2:
cc @daixiang0

artursouza (commented):
LGTM. Please, address comment from @daixiang0

logger/empty_logger.go
daixiang0: @1046102779

1046102779:
> LGTM. Please, address comment from @daixiang0

Done.

logger/default.go
```diff
@@ -0,0 +1,60 @@
+package logger
```
ItalyPaleAle, line 1: Please add the Dapr copyright header

logger/default.go
```diff
@@ -0,0 +1,60 @@
...
+
+import "io"
+
+type defaultLogger struct{}
```
ItalyPaleAle, line 5: Can you rename this to something that makes it clearer that it's a no-op logger? For example `nopLogger`?

logger/default.go
```diff
@@ -0,0 +1,60 @@
...
+func (d *defaultLogger) IsOutputLevelEnabled(level LogLevel) bool { return true }
+
+// WithLogType specify the log_type field in log. defaultLogger value is LogTypeLog.
+func (d *defaultLogger) WithLogType(logType string) Logger {
+	return nil
+}
+
+// WithFields returns a logger with the added structured fields.
+func (d *defaultLogger) WithFields(fields map[string]any) Logger {
+	return nil
```
ItalyPaleAle, line 29: `return d` or consumers may panic

logger/default.go
```diff
@@ -0,0 +1,60 @@
...
+
+// SetOutput sets the destination for the logs
+func (d *defaultLogger) SetOutput(dst io.Writer) {}
+
+// IsOutputLevelEnabled returns true if the logger will output this LogLevel.
+func (d *defaultLogger) IsOutputLevelEnabled(level LogLevel) bool { return true }
+
+// WithLogType specify the log_type field in log. defaultLogger value is LogTypeLog.
+func (d *defaultLogger) WithLogType(logType string) Logger {
+	return nil
```
ItalyPaleAle, line 24: `return d` or consumers may panic

logger/logger.go
```diff
@@ -150,3 +155,19 @@ func getLoggers() map[string]Logger {
 
 	return l
 }
+
+// NewContext returns a new Context, derived from ctx, which carries the
+// provided Logger.
+func NewContext(ctx context.Context, logger Logger) context.Context {
+	return context.WithValue(ctx, contextKey{}, logger)
```
ItalyPaleAle, line 162: I believe it's best to use a constant value rather than re-creating it every time. So:

```go
type logContextKeyType struct{}

// logContextKey is how we find Loggers in a context.Context
var logContextKey = logContextKeyType{}

// ... down below

return context.WithValue(ctx, logContextKey, logger)
```

ItalyPaleAle (changes requested):

1046102779:
Done.

 Thanks your advise @ItalyPaleAle .

ItalyPaleAle (approved):

PR #25, opened 2022-10-24
Added SetOutput method to logger

ItalyPaleAle:
This is particularly useful when building tests


artursouza (approved):

codefromthecrypt:
can you tag a release so we can use this in components-contrib?

PR #26, opened 2022-10-28
Bring in fswatcher from runtime

ItalyPaleAle:
This brings the [fswatcher](https://github.com/dapr/dapr/tree/master/pkg/fswatcher) package from the runtime repo so it can be used in components-contrib too.

As the package is brought in, it now has some improvements:

- Better batching, so only one notification is sent after a batch of changes and not multiple
- Better error handling
- Added unit tests


mukundansundar (approved):

fswatcher/watch.go
```diff
@@ -0,0 +1,59 @@
...
+
+	batchCh := make(chan struct{}, 1)
+	defer close(batchCh)
+
+	for {
+		select {
+		// Watch for events
+		case event := <-watcher.Events:
+			if event.Op&fsnotify.Create == fsnotify.Create ||
+				event.Op&fsnotify.Write == fsnotify.Write {
```
shubham1172, line 35: Should we also add support for `Remove`, `Rename` and `Chmod`? If not, we should add it to the documentation comment.

fswatcher/watch.go
```diff
@@ -0,0 +1,59 @@
...
+				if strings.Contains(event.Name, dir) {
+					// Batch the change
+					select {
+					case batchCh <- struct{}{}:
+						go func() {
+							time.Sleep(500 * time.Millisecond)
+							eventCh <- struct{}{}
```
shubham1172, line 42: Should we also read from `batchCh` here to "reset" it? Since `batchCh` has a buffer size of 1, once it gets a `struct{}{}`, it will always go into `default`, even after the notification after 500ms is sent to `eventCh`.

fswatcher/watch.go
ItalyPaleAle: They were not in the code that was originally in the runtime and for the way it's used in the runtime, I don't think they're needed.

fswatcher/watch.go
ItalyPaleAle: Yes, you're right. Thanks for catching that, will fix it

ItalyPaleAle: Fixed

mukundansundar (approved):

shubham1172 (approved):

fswatcher/watch.go
mukundansundar: @ItalyPaleAle I know that this PR is just bringing what is existing in runtime into kit repo. But I believe it will be better if we can have some form of struct detailing the event being sent back in the `eventCh`. This will later allow for watcher to be extended without extensive modifications to runtime. 
For example watcher is being used in https://github.com/dapr/dapr/pull/5197, I was also thinking we can use it in CLI to watch for file changes and restart dapr or application as a improvement for dev experience --- 

My thought was in the `eventCh`, since its a batch, we can pass in a list of events, essentially whether its a create or a write and the corresponding file details... That would be helpful on the caller side ... 
thoughts ?

mukundansundar: On other note , this can be a separate change also ...

ItalyPaleAle: That’s a good suggestion but I wonder if in that case you’d be best just using `fsnotify/fsnotify` instead?

This package is used in the runtime to watch for TLS certs, where there are 2 files that are changed at the same time. The runtime doesn’t really need the complexity of knowing which files have changed, and batching is required.

shubham1172 (approved):

artursouza (approved):

PR #35, opened 2022-12-06
Port time utils from dapr/dapr

ItalyPaleAle:
This PR ports time utils from dapr/dapr (specifically [pkg/actors/actors.go](https://github.com/dapr/dapr/blob/01891417063587e34b77c149eef5b9aae592fac4/pkg/actors/actors.go#L2105)) into dapr/kit. No code changes, just feels like this is a better place for these utils rather than being buried at the end of a file in pkg/actors.

PS: Also updates the version of the linter to 1.50.1, the same one we are currently using on other Dapr repos

### DIFF
--- a/.github/workflows/kit.yml
+++ b/.github/workflows/kit.yml
@@ -25,7 +25,7 @@ jobs:
       GOOS: ${{ matrix.target_os }}
       GOARCH: ${{ matrix.target_arch }}
       GOPROXY: https://proxy.golang.org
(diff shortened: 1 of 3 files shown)


johnewart (dismissed):

artursouza (approved):

ItalyPaleAle:
The failing tests on Windows seem unrelated and #33 fixes those

PR #42, opened 2023-03-29
Dapr crypto scheme v1 (high level methods)

ItalyPaleAle:
# Description

This PR implements the Dapr crypto scheme v1, as defined in dapr/proposals#3 

It includes the methods `Encrypt` and `Decrypt` which are high-level methods to encrypt and decrypt messages of arbitrary length (up to 256TB). These methods work on streams and don't hold more than 64KB of data in a buffer.

This PR includes extensive unit tests that cover all edge cases.

## Example usage

To encrypt a message, you invoke the `Encrypt` method with a readable stream containing the plaintext data and an object containing options.

The most important option is `WrapKeyFn`, which is a function that, given a plaintext file key (randomly-generated by the `Encrypt` method), the name of a key in the vault, and the algorithm, returns the wrapped key. Callers are required to pass this method as a lambda, which will perform wrapping using a component (for example, Azure Key Vault).

```go
enc, err := Encrypt(
    bytes.NewReader(message),
    EncryptOptions{
        WrapKeyFn: func(wrappedKey []byte, algorithm, keyName string, nonce, tag []byte) (plaintextKey jwk.Key, err error) {
            //…
        },
        KeyName:   keyName,
        Algorithm: algorithm,
    },
)
```

The method returns a readable stream `enc` which will have the data written to it.

The method does not accept a context because it is not needed when working with streams. Callers can stop the method by stopping the input stream. The `WrapKeyFn` should manage its context (for the network call) internally (i.e. inside the lambda).

The signature of the `Decrypt` method is similar, except it takes less options, with the only required one being an `UnwrapKeyFn` (same concept as key wrapping for `Encrypt`):

```go
dec, err := Decrypt(
    enc,
    DecryptOptions{
        UnwrapKeyFn: unwrapKeyFn,
    },
)
```

Here too, `enc` is a readable stream to the encrypted document, and the returned `dec` is a readable stream containing the plaintext.

## Issue reference

Tracked by: dapr/dapr#6147
Proposal: dapr/proposals#3

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


crypto/schemes/enc/v1/README.md
```diff
@@ -0,0 +1,137 @@
...
+```text
+header || binary_payload
+```
+
+## Header
+
+The **header** is human-readable and contains 3 items, each terminated by a line feed (`0x0A`) character:
+
+1. Name and version of the encryption scheme used. In this version of the spec, this is always `dapr.io/enc/v1`.
+2. The manifest, which is a JSON object.
```
tmacam, line 35: I think there is an implicit requirement that line feed characters (`0x0A`) are not part of the serialized JSON  output. [RFC 8259 - The JavaScript Object Notation (JSON) Data Interchange Format - Section 2](https://www.rfc-editor.org/rfc/rfc8259#section-2) mentions line feed as an allowed insignificant white space

> Insignificant whitespace is allowed before or after any of the six structural characters.

Something like

```json
{"k":"mykey","kw":1,"wfk":"hGYjwDpWEXEymSTFZ95zgX8krElb3Gqyls67R8zJA3k=","cph":1,"np":"Y3J5cHRvIQ==",
"this line break just before this key using a \u000A will break our parsing" :
"but it is valid JSON"}
``` 


```suggestion
2. The manifest, which is a JSON object with insignificant whitespace removed.
```

crypto/schemes/enc/v1/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+```text
+dapr.io/enc/v1
+{"k":"mykey","kw":1,"wfk":"hGYjwDpWEXEymSTFZ95zgX8krElb3Gqyls67R8zJA3k=","cph":1,"np":"Y3J5cHRvIQ=="}
+pBDKLrhAWL7IAvDKBV/v7lmbTG6AEZbf3srUN0Pnn30=
+```
+
+### Manifest
+
+The second line in the header is the **manifest**, which is a compact JSON object.
```
tmacam, line 50: This ties back to my past comment: do we have any definition of what is _"compact JSON"_?

crypto/schemes/enc/v1/algorithms.go
```diff
@@ -0,0 +1,99 @@
...
+const (
+	KeyAlgorithmAESKW      KeyAlgorithm = "AES-KW"
+	KeyAlgorithmRSAOAEP256 KeyAlgorithm = "RSA-OAEP-256"
+
+	KeyAlgorithmAES KeyAlgorithm = "AES" // Alias for AES-KW
+	KeyAlgorithmRSA KeyAlgorithm = "RSA" // Alias for RSA-OAEP-256
+)
+
+// Validate the passed algorithm and resolves aliases.
+func (a KeyAlgorithm) Validate() (KeyAlgorithm, error) {
```
tmacam, line 33: Stopped here

crypto/schemes/enc/v1/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+Its corresponding Go struct is:
+
+```go
+type Manifest struct {
+	// Name of the key that can be used to decrypt the message.
+	// This is optional, and if specified can be in the format `key` or `key/version`.
+	KeyName string `json:"k,omitempty"`
+	// ID of the wrapping algorithm used.
```
tmacam, line 61: We do not define 0x00 here or for WFK. Is this intentional to leave anything else as undefined?

crypto/schemes/enc/v1/algorithms.go
```diff
@@ -0,0 +1,99 @@
...
+
+// ID returns the numeric ID for the algorithm.
+func (a KeyAlgorithm) ID() int {
+	switch a {
+	case KeyAlgorithmAESKW, KeyAlgorithmAES:
+		return 1
+	case KeyAlgorithmRSAOAEP256, KeyAlgorithmRSA:
+		return 2
+	default:
+		return 0
```
tmacam, line 60: Tying back to the comment about we don't define what 0x00 is for this field. Should this return an error instead?

crypto/schemes/enc/v1/README.md
ItalyPaleAle:
```suggestion
2. The manifest, which is a JSON object, compacted with all the unnecessary whitespace removed.  
   Note that per the JSON specs, newline characters within a string are encoded as the string `\n`, which would not interpreted as newlines while scanning the manifest.
```

crypto/schemes/enc/v1/README.md
ItalyPaleAle:
```suggestion
The second line in the header is the **manifest**, which is a compact JSON object (i.e. with all the unnecessary whitespace removed).

> Note that per the JSON spec, newline characters within a string are safely encoded as the string `\n`, which would not interpreted as newlines while scanning the manifest.
```

crypto/schemes/enc/v1/algorithms.go
ItalyPaleAle: Yes this is intentional, and just needed a default value.

Any other algorithm would fail when `Manifest.Validate()` is called. And although we could return an error here, it seemed easier to handle a single return code.

crypto/schemes/enc/v1/ciphers.go
```diff
@@ -0,0 +1,88 @@
...
+
+// Validate the passed cipher and resolves aliases.
+func (a Cipher) Validate() (Cipher, error) {
+	switch a {
+	// Valid ciphers, not aliased
+	case CipherAESGCM, CipherChaCha20Poly1305:
+		return a, nil
+
+	default:
+		return a, errors.New("cipher " + string(a) + " is not supported")
```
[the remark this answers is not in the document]
ItalyPaleAle: Same reason, let's keep the value

crypto/schemes/enc/v1/README.md
ItalyPaleAle: Sorry I missed this.

Not defining value `0` is on purpose, so if that were to be encountered it would be considered an error. Otherwise, it'd be harder to distinguish value missing vs intentionally setting this to the zero value.

crypto/schemes/enc/v1/README.md
```diff
@@ -0,0 +1,146 @@
...
+```text
+header || binary_payload
+```
+
+## Header
+
+The **header** is human-readable and contains 3 items, each terminated by a line feed (`0x0A`) character:
+
+1. Name and version of the encryption scheme used. In this version of the spec, this is always `dapr.io/enc/v1`.
+2. The manifest, which is a JSON object, compacted with all the unnecessary whitespaces removed.
```
tmacam, line 39: Might be interesting to link to JSON RFC as it defines what __unnecessary whitespace__. I know that for go this is a non-issue as that seems to be the default for `encoding/json.Marshall`. Ditto for nodejs. But that isn't the default for [python]( https://stackoverflow.com/questions/16311562/python-json-without-whitespaces) AFAIK.

crypto/schemes/enc/v1/README.md
```diff
@@ -0,0 +1,146 @@
...
+dapr.io/enc/v1
+{"k":"mykey","kw":1,"wfk":"hGYjwDpWEXEymSTFZ95zgX8krElb3Gqyls67R8zJA3k=","cph":1,"np":"Y3J5cHRvIQ=="}
+pBDKLrhAWL7IAvDKBV/v7lmbTG6AEZbf3srUN0Pnn30=
+```
+
+### Manifest
+
+The second line in the header is the **manifest**, which is a compact JSON object (i.e. with all the unnecessary whitespaces removed).
+
+> Note that per the JSON spec, newline characters within a string are safely encoded as the string `\n`, which would not interpreted as newlines while scanning the manifest.
```
tmacam, line 56: Might be interesting to link to JSON RFC as it defines what __unnecessary whitespace__. I know that for go this is a non-issue as that seems to be the default for `encoding/json.Marshall`. Ditto for nodejs. But that isn't the default for [python]( https://stackoverflow.com/questions/16311562/python-json-without-whitespaces) AFAIK.

crypto/schemes/enc/v1/filekey.go
```diff
@@ -0,0 +1,270 @@
...
+	// Key used to encrypt the payload
+	payloadKey []byte
+}
+
+func newFileKey(cipher Cipher) (fileKey, error) {
+	// Read 39 random bytes for the file key (256 bits) and nonce prefix (56 bits)
+	rnd := make([]byte, 39)
+	_, err := io.ReadFull(rand.Reader, rnd)
+	if err != nil {
+		return fileKey{}, fmt.Errorf("failed to generate file key: %w", err)
```
tmacam, line 52: Go-newbie question: why are you returning an empty `fileKey` instead of just returning `nil`. Same on `importFileKey`

ItalyPaleAle: That method returns a non-pointer value, so cannot return `nil` there.

mukundansundar:
@ItalyPaleAle there seems to be a flaky test in here. Can you take a look?

ItalyPaleAle:
@mukundansundar looks like there was a bug in the crypto code but not in what was added with this PR. Fixed here: https://github.com/dapr/kit/pull/44

JoshVanL (changes requested):

JoshVanL:
@ItalyPaleAle not completely finished the _code_ part of the review, but thought it was worth sending my comments about the design early :slightly_smiling_face:

schemes/enc/v1/README.md
```diff
@@ -0,0 +1,146 @@
+# Dapr encryption scheme v1: `dapr.io/enc/v1`
```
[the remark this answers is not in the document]
yaron2: Indeed, please use `encryption`.

schemes/enc/v1/README.md
```diff
@@ -0,0 +1,146 @@
...
+The FK is wrapped using a key stored in a key vault (**Key Encryption Key (KEK)**) by Dapr. The result of the wrapping operation is the **Wrapped File Key (WFK)**. The algorithm used depends on the type of the KEK as well as the algorithms supported by the component:
+
+- For symmetric keys:
+    - AES-KW with 256-bit keys ([RFC 3394](https://www.rfc-editor.org/rfc/rfc3394.html)): `A256KW`
+		- Because the File Key is 256-bit long, only 256-bit wrapping keys can be used
+	- AES-CBC with keys 128-bit, 192-bit, and 256-bit: `A128CBC-NOPAD`, `A192CBC-NOPAD`, `A256CBC-NOPAD`
+		- These don't use PKCS#7 padding because the File Key is 256-bit so it's a multiple of the AES block size.
+- For RSA keys:
+    - RSA OAEP with SHA-256: `RSA-OAEP-256`
+		- Dapr doesn't impose limitations on the size of the key, and any key bigger than 1024 bits should work; however, 4096-bit keys are strongly recommended.
```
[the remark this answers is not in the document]
ItalyPaleAle: It's a recommendation only, and we won't enforce it.

However, 2048 is probably not enough these days anymore, as it has only ~112 bits of security.

schemes/enc/v1/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+> Note that per the [JSON spec (RFC 8259)](https://www.rfc-editor.org/rfc/rfc8259), newline characters within a string are safely encoded as the string `\n`, which would not interpreted as newlines while scanning the manifest. The spec also defines what constitutes "unnecessary whitespace".
+
+Its corresponding Go struct is:
+
+```go
+type Manifest struct {
+	// Name of the key that can be used to decrypt the message.
+	// This is optional, and if specified can be in the format `key` or `key/version`.
+	KeyName string `json:"k,omitempty"`
```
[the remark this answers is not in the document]
ItalyPaleAle: Since this is included in each encrypted message, i'd go for conciseness to save some bytes. It's not expected to be read by humans anyways.

schemes/enc/v1/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+```text
+dapr.io/enc/v1
+{"k":"mykey","kw":1,"wfk":"hGYjwDpWEXEymSTFZ95zgX8krElb3Gqyls67R8zJA3k=","cph":1,"np":"Y3J5cHRvIQ=="}
+pBDKLrhAWL7IAvDKBV/v7lmbTG6AEZbf3srUN0Pnn30=
+```
+
+### Manifest
+
+The second line in the header is the **manifest**, which is a compact JSON object (i.e. with all the unnecessary whitespaces removed).
```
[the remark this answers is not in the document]
ItalyPaleAle: I will remove that.

schemes/enc/v1/README.md
```diff
@@ -0,0 +1,146 @@
...
+```text
+header || binary_payload
+```
+
+## Header
+
+The **header** is human-readable and contains 3 items, each terminated by a line feed (`0x0A`) character:
+
+1. Name and version of the encryption scheme used. In this version of the spec, this is always `dapr.io/enc/v1`.
+2. The manifest, which is a JSON object, compacted with all the unnecessary whitespaces removed.
```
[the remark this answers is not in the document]
ItalyPaleAle: The idea is:

- First line indicates the version. We shouldn't try to parse the rest of the document is the version mismatches. If we put the version in the JSON, then the version may not be at the beginning and we could waste time parsing the entire header.
- Second like is the JSON manifest
- Third line is the MAC. This can't be within the manifest for obvious reasons.

If the JSON is compacted, there shouldn't be any newlines characters (as in `0xA` bytes). Within a JSON strings, newlines are required to be encoded as `"\n"`. I don't expect that we will need to put freeform text in the manifest anyways, so this is just us being overly cautious.

schemes/enc/v1/README.md
```diff
@@ -0,0 +1,146 @@
...
+	// Name of the key that can be used to decrypt the message.
+	// This is optional, and if specified can be in the format `key` or `key/version`.
+	KeyName string `json:"k,omitempty"`
+	// ID of the wrapping algorithm used.
+	// 0x01 = A256KW
+	// 0x02 = A128CBC-NOPAD
+	// 0x03 = A192CBC-NOPAD
+	// 0x04 = A256CBC-NOPAD
+	// 0x05 = RSA-OAEP-256
+	KeyWrappingAlgorithm int `json:"kw"`
```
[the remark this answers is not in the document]
ItalyPaleAle: Since this is not meant for human consumption, I would go with saving some bytes since this is included in the encrypted document.

schemes/enc/v1/algorithms.go
```diff
@@ -0,0 +1,116 @@
...
+	case KeyAlgorithmAES128CBC:
+		return 2
+	case KeyAlgorithmAES192CBC:
+		return 3
+	case KeyAlgorithmAES256CBC:
+		return 4
+	case KeyAlgorithmRSAOAEP256, KeyAlgorithmRSA:
+		return 5
+	default:
+		return 0
```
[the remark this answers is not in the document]
ItalyPaleAle: I will keep this to `0` because it's not a valid algorithm and it's the zero value when unmarshalling an empty field into the struct.

schemes/enc/v1/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+```text
+header || binary_payload
+```
+
+## Header
+
+The **header** is human-readable and contains 3 items, each terminated by a line feed (`0x0A`) character:
+
+1. Name and version of the encryption scheme used. In this version of the spec, this is always `dapr.io/enc/v1`.
```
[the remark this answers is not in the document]
ItalyPaleAle: Although `enc.dapr.io/v1` makes sense, I would rather not change this now unless there's a strong reason for doing that, as per my other comment.

schemes/enc/v1/README.md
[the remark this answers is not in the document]
artursouza: If this is not for human consumption, then why the nicely formatted domain? "Enc" is too ambiguous. It can mean encryption or encoding.

artursouza:
I recommend not marking comments as resolved unless the reviewer agrees or is unresponsive.

schemes/enc/v1/README.md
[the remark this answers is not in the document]
ItalyPaleAle: > If this is not for human consumption, then why the nicely formatted domain? 

It is common pattern to make them URLs. For example, age, which was a big inspiration for this scheme, does that too: https://github.com/C2SP/C2SP/blob/main/age.md#header

The benefit is that it's a valid URL, so I've already opened a [PR in dapr/website](https://github.com/dapr/website/pull/76) to make that point to the README with the specs. If someone were to read that and had doubts on the meaning of "enc", they could visit the URL.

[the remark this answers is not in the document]
yaron2: > They're not meant for human consumption

Contributors consume this code when, don't they?

> For example, age, which was a big inspiration for this scheme, does that too: https://github.com/C2SP/C2SP/blob/main/age.md#header

Age is not our concern.

> so less bytes in the output files, the better.

The benefit of creating clarity far outweighs the negligible and vastly insignificant increase of 7 bytes. Had length been an issue I'd imagine you wouldn't have constructed this is a URL to begin with.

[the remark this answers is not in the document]
ItalyPaleAle: As discussed on discord this will not change in this PR.

after the PR is merged, changes are welcome to the URL if anyone were up for that. They would likely require changes to Dapr/dapr, go-sdk, website, and possibly docs.

schemes/enc/v1/algorithms.go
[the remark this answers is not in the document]
daixiang0: From `NewKeyAlgorithmFromID`, `0` is invalid neither. Maybe a invalid const like `InvalidID`?

schemes/enc/v1/ciphers.go
```diff
@@ -0,0 +1,91 @@
...
+const (
+	CipherAESGCM           Cipher = "AES-GCM"
+	CipherChaCha20Poly1305 Cipher = "CHACHA20-POLY1305"
+
+	cipherNumAESGCM = 1
+	cipherNumChaCha20Poly1305 = 2
+)
+
+// Validate the passed cipher and resolves aliases.
+func (a Cipher) Validate() (Cipher, error) {
```
daixiang0, line 33: I would like `func (c Cipher) Validate() (Cipher, error) {`, `a` looks meaningless.

schemes/enc/v1/ciphers.go
```diff
@@ -0,0 +1,91 @@
...
+	switch a {
+	// Valid ciphers, not aliased
+	case CipherAESGCM, CipherChaCha20Poly1305:
+		return a, nil
+
+	default:
+		return a, errors.New("cipher " + string(a) + " is not supported")
```
daixiang0, line 40: Could we use `fmt` rather than `+`?

schemes/enc/v1/manifest.go
```diff
@@ -0,0 +1,55 @@
...
+		return fmt.Errorf("key wrapping algorithm is invalid: %w", err)
+	}
+	if len(m.WFK) == 0 {
+		return errors.New("wrapped file key is empty")
+	}
+	m.Cipher, err = m.Cipher.Validate()
+	if err != nil {
+		return fmt.Errorf("cipher is invalid: %w", err)
+	}
+	if len(m.NoncePrefix) != 7 {
```
daixiang0, line 50: magic number

schemes/enc/v1/ciphers.go
ItalyPaleAle: Done

schemes/enc/v1/ciphers.go
ItalyPaleAle: Sure, it's an error here anyways so (hopefully) not a hot path

schemes/enc/v1/algorithms.go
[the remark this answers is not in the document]
ItalyPaleAle: Done

schemes/enc/v1/manifest.go
ItalyPaleAle: Added constant

schemes/enc/v1/README.md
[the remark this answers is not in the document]
artursouza: We agreed to keep it as-is for now. We might decide to make it a breaking change in the future but is not a blocker for this PR.

artursouza (approved):

schemes/enc/v1/README.md
```diff
@@ -0,0 +1,146 @@
...
+	// 0x02 = ChaCha20-Poly1305
+	Cipher int `json:"cph"`
+	// Random sequence of 7 bytes generated by a CSPRNG.
+	NoncePrefix []byte `json:"np"`
+}
+```
+
+- **`KeyName`** is the name of the key that can be used to decrypt the message.  
+  Usually this is the same as the name of the key used to encrypt the message, but when asymmetric ciphers are used, it could be different.  
+  Including a `KeyName` in the manifest is not required, but when i'ts present, it's used as the default value for the key name while decrypting the document (however, users can override this value by passing a custom one while decrypting the document).
```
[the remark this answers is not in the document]
artursouza: We need to keep this moving. Josh is out in KubeCon EU, so we will merge this PR. This is an Alpha feature and can have breaking changes.

schemes/enc/v1/README.md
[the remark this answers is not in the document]
artursouza: Josh is out in KubeCon and we can revisit this since it is an Alpha feature and can afford a breaking change.

PR #46, opened 2023-04-18
Add jwkscache package

ItalyPaleAle:
Adds the "jwkscache" package, which contains a struct that can be used to create a cache of a JWKS (JWK Set). 

It supports retrieving a JWKS from:

- A path on the local disk. This is watched with fsnotify to automatically reload the JWKS when the file changes on disk.
- A HTTP(S) URL. This is automatically refreshed if a caller requests a key that isn't in the cached set.
- A JWKS passed during initialization, optionally base64-encoded.

This is currently in use by the [JWKS crypto component](https://github.com/dapr/components-contrib/blob/0221ad7edff3f33cb28fa6c7e6d8dcaee44bd5fb/crypto/jwks/component.go) (code has been extrapolated from there, and it now includes some bug fixes and the addition of tests). We are going to need it in the runtime too for Sentry (see work going on in dapr/dapr#6171)

### DIFF
--- a/jwkscache/cache.go
+++ b/jwkscache/cache.go
@@ -1,0 +1,267 @@
+/*
+Copyright 2023 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package jwkscache contains utils to manage a cache of a JWK Set (via jwk.Set).
+// It supports retrieving a JWKS from:
+//
+// - A path on the local disk. This is watched with fsnotify to automatically reload the JWKS when the file changes on disk.
+// - A HTTP(S) URL. This is automatically refreshed if a caller requests a key that isn't in the cached set.
+// - A JWKS passed during initialization, optionally base64-encoded.
+package jwkscache
+
+import (
+	"context"
+	"crypto/tls"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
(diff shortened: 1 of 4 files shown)


jwkscache/cache.go
```diff
@@ -0,0 +1,267 @@
...
+		// If we don't get a response in 5s, something bad's going on
+		return errors.New("failed to initialize JWKS from file: no file loaded after 5s")
+	case <-ctx.Done():
+		return fmt.Errorf("failed to initialize JWKS from file: %w", ctx.Err())
+	}
+}
+
+// Used by initJWKSFromFile to parse a JWKS file every time it's changed
+func (c *JWKSCache) parseJWKSFile(file string) error {
+	c.logger.Debugf("Reloading JWKS file from disk")
```
daixiang0, line 250: The first time is not "reloading", "parsing" is better.

ItalyPaleAle: Done

ItalyPaleAle: (I have moved it a few lines above where we have `firstDone`)

jwkscache/cache.go
```diff
@@ -0,0 +1,270 @@
...
+				}
+				err := c.parseJWKSFile(file)
+				if !firstDone {
+					// The first time, signal that the initialization was complete and pass the error
+					loaded <- err
+					close(loaded)
+					firstDone = true
+				} else if err != nil {
+					// Log errors only
+					c.logger.Errorf("Error reading JWKS from disk: %v", err)
```
daixiang0, line 230: How about load file when init and keep monitor about file changes? If so, we can remove `firstDone`.

ItalyPaleAle: Doing it this way allows us to remove race conditions because there's only 1 goroutine that is loading the file. If we also load the file in the "main" goroutine, there's a possibility of a race condition if the file is changed on disk at the exact same time.

ItalyPaleAle: At the same time, we also cannot start the watcher **after** having read the file in the "main" goroutine. There are no race conditions there, but in that case if the file changes on disk after we read it, but before the background goroutine is started, we may miss the updated file.

daixiang0: Good explanation, please add comments here.

daixiang0 (approved):
Please create an issue to update jwks component.

ItalyPaleAle:
> Please create an issue to update jwks component.

I will have the PR done tomorrow in contrib

PR #54, opened 2023-08-02
Added eventqueue package

ItalyPaleAle:
Add the generic "eventqueue" package, which is used in dapr/dapr#6716 and will be used in other PRs in dapr/dapr. This code is generic and not specific to Dapr, so dapr/kit seems a good place for it.

Eventqueue maintains an in-memory queue of messages that have to be "executed" at a given time in the future. The queue is a priority queue where messages are in the order they are executed. More details are in dapr/dapr#6716

Users interact with the Processor object which processes messages from the queue at the given time, efficiently using at most one goroutine that is blocked on the next mesage.


eventqueue/processor.go
```diff
@@ -0,0 +1,229 @@
...
+			// Use a select in case another goroutine is sending a reset signal too
+			select {
+			case p.resetCh <- struct{}{}:
+			default:
+			}
+		}
+		return
+	}
+
+	go p.processLoop()
```
JoshVanL, line 138: This should have a wait group to close can wait on this routine to return.

eventqueue/processor.go
```diff
@@ -0,0 +1,229 @@
...
+	if !p.stopped.CompareAndSwap(false, true) {
+		// Already stopped
+		return nil
+	}
+
+	// Send a signal to stop
+	close(p.stopCh)
+
+	// Blocks until process loop ends
+	t := time.NewTimer(5 * time.Second)
```
JoshVanL, line 104: Use `time.After`.

eventqueue/processor.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	p.queueLock.Unlock()
+
+	return nil
+}
+
+// Stop the processor.
+func (p *Processor[T]) Close() error {
+	if !p.stopped.CompareAndSwap(false, true) {
+		// Already stopped
```
JoshVanL, line 96: We should move the close channel to this if block, and always wait for the processor to finishing running when calling Close.

eventqueue/processor.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// We need to check if this is the next item in the queue, as that requires stopping the processor
+	p.queueLock.Lock()
+	peek, ok := p.queue.Peek()
+	p.queue.Remove(key)
+	if ok && peek.Key() == key {
+		// If the item was the first one in the queue, restart the processor
+		p.process(true)
+	}
+	p.queueLock.Unlock()
```
JoshVanL, line 88: Please always defer the unlock unless for a specific reason.

eventqueue/processor.go
```diff
@@ -0,0 +1,229 @@
...
+	return nil
+}
+
+// Dequeue removes a item from the queue.
+func (p *Processor[T]) Dequeue(key string) error {
+	if p.stopped.Load() {
+		return ErrProcessorStopped
+	}
+
+	// We need to check if this is the next item in the queue, as that requires stopping the processor
```
JoshVanL, line 80: Why do we need to stop the processor?

eventqueue/queue.go
```diff
@@ -0,0 +1,163 @@
...
+// Note: methods in this struct are not safe for concurrent use. Callers should use locks to ensure consistency.
+type queue[T queueable] struct {
+	heap  *queueHeap[T]
+	items map[string]*queueItem[T]
+}
+
+// newQueue creates a new queue.
+func newQueue[T queueable]() *queue[T] {
+	return &queue[T]{
+		heap:  &queueHeap[T]{},
```
JoshVanL, line 40: but

```suggestion
		heap:  new(queueHeap[T]),
```

JoshVanL (dismissed):
Some comments from me :slightly_smiling_face:

eventqueue/processor.go
```diff
@@ -0,0 +1,229 @@
...
+
+package eventqueue
+
+import (
+	"errors"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	kclock "k8s.io/utils/clock"
```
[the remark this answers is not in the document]
ItalyPaleAle: That will be shadowed by the variables named "clock". We also normally import this package as "kclock".

eventqueue/processor.go
ItalyPaleAle: time.After cannot be garbage-collected until it fires. In this case, 5s is a relatively long time, and with time.NewTimer we can stop it earlier and mark it as garbage sooner.

From [docs](https://pkg.go.dev/time#After):

> After waits for the duration to elapse and then sends the current time on the returned channel. It is equivalent to NewTimer(d).C. The underlying Timer is not recovered by the garbage collector until the timer fires. If efficiency is a concern, use NewTimer instead and call Timer.Stop if the timer is no longer needed.

eventqueue/processor.go
```diff
@@ -0,0 +1,229 @@
...
+
+// ErrProcessorStopped is returned when the processor is not running.
+var ErrProcessorStopped = errors.New("processor is stopped")
+
+// Processor manages the queue of items and processes them at the correct time.
+type Processor[T queueable] struct {
+	executeFn          func(r T)
+	queue              *queue[T]
+	clock              kclock.Clock
+	queueLock          sync.Mutex
```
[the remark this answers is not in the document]
ItalyPaleAle: Since it's used for the queue only, I like `queueLock` which is more specific.

eventqueue/processor.go
ItalyPaleAle: To avoid having a goroutine running if there's nothing in the queue and wasting resources. Otherwise the test would complain we added a goroutine that is always running :)

eventqueue/processor.go
ItalyPaleAle: Here it should not make much difference, but I like the explicitness of seeing when the lock is released.

Defer is a bit like a "goto" and alters the natural flow of a program. Although it's been a while since I was in high school, the memory of my CS teacher repeating pretty much every week how maintaining the natural flow of a program is important and "goto"'s are bad are hard to get rid of :)

eventqueue/processor.go
ItalyPaleAle: To the block where there's currently the "Already stopped" comment? But then we may close the channel twice which would cause a panic?

eventqueue/processor.go
```diff
@@ -0,0 +1,229 @@
...
+	// Blocks until process loop ends
+	t := time.NewTimer(5 * time.Second)
+	select {
+	case p.processorRunningCh <- struct{}{}:
+		// All good
+		if !t.Stop() {
+			<-t.C
+		}
+	case <-t.C:
+		return errors.New("Processor loop did not stop in 5s")
```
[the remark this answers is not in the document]
ItalyPaleAle: I agree, this makes sense. I'll change this.

eventqueue/processor.go
```diff
@@ -0,0 +1,229 @@
...
+		// If we get a reset signal, restart the loop
+		case <-p.resetCh:
+			// Restart the loop
+			continue
+
+		// If we receive a stop signal, exit
+		case <-p.stopCh:
+			// Stop the timer and exit the loop
+			if !t.Stop() {
+				<-t.C()
```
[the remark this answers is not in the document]
ItalyPaleAle: From docs: https://pkg.go.dev/time#Timer.Stop

eventqueue/processor.go
```diff
@@ -0,0 +1,229 @@
...
+// Executes a item when it's time.
+func (p *Processor[T]) execute(r T) {
+	// Pop the item now that we're ready to process it
+	// There's a small chance this is a different item than the one we peeked before
+	p.queueLock.Lock()
+	// For safety, let's peek at the first item before popping it and make sure it's the same object
+	// It's unlikely, but if it's a different object then restart the loop
+	peek, ok := p.queue.Peek()
+	if !ok || peek != r {
+		p.queueLock.Unlock()
```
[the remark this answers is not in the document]
ItalyPaleAle: I'd rather have explicit controls on the locks here so I can release them as soon as they're not needed anymore, since they block every other operation.

eventqueue/processor.go
ItalyPaleAle: Changed. Although we don't need a wg because we have `processorRunningCh`

eventqueue/processor.go
```diff
@@ -0,0 +1,229 @@
...
+		default:
+			// Nop, proceed
+		}
+
+		scheduledTime = r.ScheduledTime()
+		deadline = scheduledTime.Sub(p.clock.Now())
+
+		// If the deadline is less than 0.5ms away, execute it right away
+		// This is more efficient than creating a timer
+		if deadline < 500*time.Microsecond {
```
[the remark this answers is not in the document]
ItalyPaleAle: In the tests I've run, without this performance was suffering a lot. This is because `time.After` (or the channel from `NewTimer`) is executed after a fairly long delay, so if something needs to be done "right away", it's best to just do this.

I mean the performance was actually noticeable: tests on actor timers were failing with a timeout before I added this.

eventqueue/processor.go
```diff
@@ -0,0 +1,229 @@
...
+	stopped            atomic.Bool
+}
+
+// NewProcessor returns a new Processor object.
+// executeFn is the callback invoked when the item is to be executed; this will be invoked in a background goroutine.
+func NewProcessor[T queueable](executeFn func(r T), clock kclock.Clock) *Processor[T] {
+	return &Processor[T]{
+		executeFn:          executeFn,
+		queue:              newQueue[T](),
+		processorRunningCh: make(chan struct{}, 1),
```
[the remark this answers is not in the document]
ItalyPaleAle: This absolutely must have a capacity of 1 because of how we're using it.

https://go.dev/play/p/TK0y6Porva7

Then try to remove the capacity and see that it ends up in a deadlock

eventqueue/queue.go
ItalyPaleAle: Changed

eventqueue/processor.go
```diff
@@ -0,0 +1,229 @@
...
+	kclock "k8s.io/utils/clock"
+)
+
+// ErrProcessorStopped is returned when the processor is not running.
+var ErrProcessorStopped = errors.New("processor is stopped")
+
+// Processor manages the queue of items and processes them at the correct time.
+type Processor[T queueable] struct {
+	executeFn          func(r T)
+	queue              *queue[T]
```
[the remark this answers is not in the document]
ItalyPaleAle: Changed

eventqueue/processor.go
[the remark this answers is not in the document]
JoshVanL: It's totally fine to call variable the same as a package, and we do it all the time. We are also using the `clock` package name more then `kclock` in `dapr/dapr`. I'd like us to keep to using `clock`.

eventqueue/processor.go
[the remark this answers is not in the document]
JoshVanL: Hm, defer is doing the exact same though right? It is executed exactly after we return except for the last case where we spawn a go routine. We are not saving anything by not using defer, and it makes things a lot harder to read/understand what the code is doing by being non-idiomatic.

eventqueue/processor.go
JoshVanL: Yes, sorry- we will need to remove the `!` operator on the if statement

eventqueue/processor.go
JoshVanL: They are nothing like gotos- there are more of a reverse function wrap imo. Using defers is the idiomatic way of unlocking a mutex in these scenarios and what the reader will be expecting. This is explicitly called out in [effective Go](https://go.dev/doc/effective_go#defer).

eventqueue/processor.go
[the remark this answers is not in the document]
ItalyPaleAle: Done

eventqueue/processor.go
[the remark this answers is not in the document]
ItalyPaleAle: image

VS Code isn't letting me do that.

eventqueue/processor.go
ItalyPaleAle: Ok should be good now.

JoshVanL: I think this is still relevant- I think `Close()` should always block until the processor has stopped running for all callers.

berndverst (approved):

PR #55, opened 2023-08-04
Remove deprecated package github.com/benbjohnson/clock

### LINKED ISSUE #50: `github.com/benbjohnson/clock` is deprecated
Package [github.com/benbjohnson/clock](https://github.com/benbjohnson/clock) is used by Dapr kit in the `cron` package, but it's been archived by the author as deprecated.

We need to replace it, possibly with `k8s.io/utils/clock` which is used in dapr/dapr for other purposes.

ItalyPaleAle:
Replaced with k8s.io/utils/clock

Fixes #50


cron/doc.go
```diff
@@ -35,9 +35,11 @@ them in their own goroutines.
 
 # Time mocking
 
-import "github.com/benbjohnson/clock"
+import (
+	clocktesting "k8s.io/utils/clock/testing"
+)
 
-clk := clock.NewMock()
+clk := clock.NewFakeClock(time.Now())
```
ItalyPaleAle, line 42:
```suggestion
clk := clocktesting.NewFakeClock(time.Now())
```

cron/cron.go
```diff
@@ -271,18 +270,25 @@ func (c *Cron) run() {
...
+			timer = c.clk.NewTimer(c.entries[0].Next.Sub(now))
+			timerCh = timer.C()
 		}
 
 		for {
 			select {
-			case now = <-timer.C:
+			case now = <-timerCh:
+				// Set timer to nil so we can exit cleanly
+				timer = nil
```
mukundansundar, line 290: why are we setting timer to nil here? 
Will the clock ticks still work if the timer reference is gone? Won't it be garbage collected?

ItalyPaleAle: If this isn’t set to nil, below on L317 we would try to call Stop() again on the timer which has already fired. This will cause a deadlock. (It’s the reason why we had https://github.com/dapr/dapr/pull/6523

mukundansundar (approved):

Issue #56, opened 2023-08-08
Dapr Error Handling/Codes - Add Internal Shared Functionality


robertojrojas:
## Describe the proposal
 Add internal shared functionality to facilitate the Error communication between the runtime and components.
This functionality will help in the implementation of the [Dapr Error Handling/Codes Proposal](https://github.com/dapr/proposals/blob/main/0009-BCIRS-error-handling-codes.md)
 As a starting point, this functionality should consists of the following:

- Custom Error(s) following the Go way:  
 https://www.digitalocean.com/community/tutorials/creating-custom-errors-in-go
 https://pkg.go.dev/errors#pkg-overview

Here is a sample Custom Error snippet:

```golang
type DaprError struct {
   err error
   code ....
   description string
   resourceInfoData *ResourceInfoData
   metadata map[string]string
}

func NewDaprError(code ...., err error) DaprError {
	return DaprError{
		code: code,
		err: err,
	}
}

// Error implements the error interface.
func (c DaprError) Error() string {
	return c.err.Error()
}

// Unwrap implements the error unwrapping interface.
func (c DaprError) Unwrap() error {
	return c.err
}

// Description returns the description of the error.
func (c DaprError) Description() string {
	if c.description != "" {
		return c.description
	}
	return c.err.Error()
}

func (c *DaprError) SetDescription(description string) {
	c.description = description
}

// Add methods to access code, resourceInfoData, metadata
// Add methods to set resourceInfoData and metadata
```    

- Utility/Helper Interfaces and Functions as needed.
- Ability to determine if the `ErrorCodes` Feature has been enabled.
- Constants to be shared between `runtime` and `components`

PR #57, opened 2023-08-09
Dapr Error Handling/Codes - Add Internal Shared Functionality

robertojrojas:
# Description

This PR adds functionality to be shared between `the Dapr runtime` and various `components`.
For more details see, issue #56 and [Dapr Error Handling/Codes Proposal](https://github.com/dapr/proposals/blob/main/0009-BCIRS-error-handling-codes.md)

## Issue reference

Please reference the issue this PR will close: #56 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


errorcodes/errorcodes.go
```diff
@@ -0,0 +1,268 @@
...
+	if _, ok := metadata[ErrorCodesFeatureMetadataKey]; ok {
+		return true
+	}
+	return false
+}
+
+// New create a new DaprError using the supplied metadata and ErrorOptions
+// **Note**: As this code is in `Feature Preview`, it will only continue processing
+// if the ErrorCodes is enabled
+func New(err error, metadata map[string]string, options ...ErrorOption) *DaprError {
```
artursouza, line 76: NewDaprError

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+// call this function to apply option
+type ErrorOption func(*DaprError)
+
+type DaprError struct {
+	err              error
+	description      string
+	reason           Reason
+	resourceInfoData *ResourceInfoData
```
artursouza, line 56: nit: just call it resourceInfo

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,275 @@
...
+var (
+	NoReason                       = Reason("NO_REASON_SPECIFIED")
+	StateETagMismatchReason        = Reason("DAPR_STATE_ETAG_MISMATCH")
+	StateETagInvalidReason         = Reason("DAPR_STATE_ETAG_INVALID")
+	TopicNotFoundReason            = Reason("DAPR_TOPIC_NOT_FOUND")
+	SecretKeyNotFoundReason        = Reason("DAPR_SECRET_KEY_NOT_FOUND")
+	ConfigurationKeyNotFoundReason = Reason("DAPR_CONFIG_KEY_NOT_FOUND")
+)
+
+type ResourceInfoData struct {
```
artursouza, line 44: ResourceInfo

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,275 @@
...
+
+type DaprError struct {
+	err              error
+	description      string
+	reason           Reason
+	resourceInfoData *ResourceInfoData
+	metadata         map[string]string
+}
+
+func ActivateErrorCodesFeature(enabled bool, metadata map[string]string) {
```
artursouza, line 60: EnableComponentErrorCode

Just take the map and let the runtime know if the feature is enabled.

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,275 @@
...
+	if enabled && metadata != nil {
+		metadata[ErrorCodesFeatureMetadataKey] = "true"
+	}
+}
+
+func featureEnabled(metadata map[string]string) bool {
```
artursouza, line 66: IsComponentErrorCodeEnabled

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,275 @@
...
+		if ese != nil {
+			return nil, ese
+		}
+		return st, nil
+	}
+
+	return nil, fmt.Errorf("unable to a DaprError from: %v", err)
+}
+
+func newStatusError(de *DaprError) (*status.Status, error) {
```
artursouza, line 183: Use it as an OO method of DaprError instead

robertojrojas:
Snippet from Component side (Redis):
```golang
	if de := kitErrorCodes.NewDaprError(err, req.Metadata, kitErrorCodes.WithReason(kitErrorCodes.StateETagMismatchReason)); de != nil {
				de.SetResourceInfoData(&r.resourceInfoData)
				de.SetDescription(fmt.Sprintf("state store Set - possible etag(%s) %s. original error: %v", *req.ETag, string(state.ETagMismatch), err))
				return de
}
```

artursouza (dismissed):

artursouza:
LGTM but needs lint fixes.

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,280 @@
...
+// EnableComponentErrorCode stores an indicator for
+// components to determine if the ErrorCodes Feature
+// is enable.
+func EnableComponentErrorCode(metadata map[string]string) {
+	if metadata != nil {
+		metadata[ErrorCodesFeatureMetadataKey] = "true"
+	}
+}
+
+func featureEnabled(metadata map[string]string) bool {
```
artursouza, line 69: If metadata can be nil above, it can be nil here too.

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,280 @@
...
+type Reason string
+
+const (
+	ErrorCodesFeatureMetadataKey = "error_codes_feature"
+	Owner                        = "components-contrib"
+	Domain                       = "dapr.io"
+)
+
+var (
+	NoReason                       = Reason("NO_REASON_SPECIFIED")
```
artursouza, line 36: Can we move the reasons out of kit? I see this list growing and it would be odd to force every new error to also have a kit change.

artursouza (commented):
Some small changes. It is close.

errorcodes/errorcodes.go
robertojrojas: Something I learned from one of the comments from other PRs about nil maps in Go is that lookup operations can be performed safely and will return the zero value of the entry value type. In this case false. Only adding an entry to a nil map is not allowed.

errorcodes/errorcodes.go
robertojrojas: That's a great point! Let me change it so only the `Reason` definition is in kit, but code depending on this functionality like `components-contrib` can add the various Reasons.

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+)
+
+var NoErrorReason = WithErrorReason(NoReason, 503, codes.Internal)
+
+type ResourceInfo struct {
+	ResourceType string
+	ResourceName string
+}
+
+func WithErrorReason(reason string, httpCode int, grpcStatusCode codes.Code) ErrorOption {
```
artursouza, line 40: Can't this be a function of DaprError instead?

artursouza (changes requested):

errorcodes/errorcodes.go
robertojrojas: The idea of the `WithErrorReason` is to provide the ability to create a `ErrorOption` values which are used as part of the DaprError construction. This is how I envision it'll be used within components-contrib
```golang
var (
	StateETagMismatchReason = errorcodes.WithErrorReason(StateStoreEtagMismatch, 400, codes.Aborted)
	StateETagInvalidReason  = errorcodes.WithErrorReason(StateStoreETagInvalid, 400, codes.Aborted)
)
...
if de := state.NewDaprError(err, req.Metadata, state.StateETagMismatchReason); de != nil {
				de.SetResourceInfo(resourceInfo)
				de.SetDescription(fmt.Sprintf("state store Set - possible etag(%s) %s. original error: %v", *req.ETag, string(state.ETagMismatch), err))
				return de
}
```

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
+/*
```
JoshVanL, line 1: Consider changing the package and directory names to be `errors` to make things cleaner. Instances of the string `Error` can then be removed from names and signatures since it is implied by the package name.

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/runtime/protoiface"
+)
+
+const (
+	ErrorCodesFeatureMetadataKey = "error_codes_feature"
```
JoshVanL, line 27: Where would this key appear?

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+	Owner                        = "components-contrib"
+	Domain                       = "dapr.io"
+	NoReason                     = "NO_REASON_SPECIFIED"
```
JoshVanL, line 30: Do these need to be exported?

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+)
+
+var NoErrorReason = WithErrorReason(NoReason, 503, codes.Internal)
```
JoshVanL, line 33: Should this be changed to `Unknown` reason?

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+
+type ResourceInfo struct {
+	ResourceType string
+	ResourceName string
```
JoshVanL, line 37: Remove type stutter.

```suggestion
	Type string
	Name string
```

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+const (
+	ErrorCodesFeatureMetadataKey = "error_codes_feature"
+	Owner                        = "components-contrib"
+	Domain                       = "dapr.io"
+	NoReason                     = "NO_REASON_SPECIFIED"
+)
+
+var NoErrorReason = WithErrorReason(NoReason, 503, codes.Internal)
+
+type ResourceInfo struct {
```
JoshVanL, line 35: Is `Info` necessary here? Can you add a comment detailing what this struct is and used for, as well as it's fields? It's not clear to me what they are for.

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+
+type DaprError struct {
+	err            error
+	description    string
+	reason         string
+	httpCode       int
+	grpcStatusCode codes.Code
+	metadata       map[string]string
+	resourceInfo   *ResourceInfo
+}
```
JoshVanL, line 60: Move to top of file

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+	f := func(er *DaprError) {
+		er.reason = reason
+		er.grpcStatusCode = grpcStatusCode
+		er.httpCode = httpCode
+	}
+	return f
+}
+
+// call this function to apply option
+type ErrorOption func(*DaprError)
```
JoshVanL, line 50: type/function declaration comments should ideally start with the name of the type or function name, i.e. `// ErrorOption applies a `DaprError` to..`

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+	reason         string
+	httpCode       int
+	grpcStatusCode codes.Code
+	metadata       map[string]string
+	resourceInfo   *ResourceInfo
+}
+
+// EnableComponentErrorCode stores an indicator for
+// components to determine if the ErrorCodes Feature
+// is enable.
```
JoshVanL, line 64: `ErrorCodesFeatureMetadataKey` should this value then have a Dapr prefix to ensure namespacing? i.e. `"_dapr_error_codes_feature"`?

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+func EnableComponentErrorCode(metadata map[string]string) {
+	if metadata != nil {
+		metadata[ErrorCodesFeatureMetadataKey] = "true"
+	}
+}
```
JoshVanL, line 69: If metadata is `nil`, wont we miss setting this feature gate on?

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+
+func featureEnabled(metadata map[string]string) bool {
+	if _, ok := metadata[ErrorCodesFeatureMetadataKey]; ok {
+		return true
+	}
+	return false
```
JoshVanL, line 75: nit:
```suggestion
	_, ok := metadata[ErrorCodesFeatureMetadataKey]
	return ok
```

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// NewDaprError create a new DaprError using the supplied metadata and ErrorOptions
+// **Note**: As this code is in `Feature Preview`, it will only continue processing
+// if the ErrorCodes is enabled
+func NewDaprError(err error, metadata map[string]string, options ...ErrorOption) *DaprError {
```
JoshVanL, line 81: Name implied by the repo and package name.

```suggestion
func New(err error, metadata map[string]string, options ...Option) *Error {
```

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+		er.grpcStatusCode = grpcStatusCode
+		er.httpCode = httpCode
+	}
+	return f
+}
+
+// call this function to apply option
+type ErrorOption func(*DaprError)
+
+type DaprError struct {
```
JoshVanL, line 52:
```suggestion
type Error struct {
```

go.mod
```diff
@@ -12,13 +12,17 @@ require (
 	github.com/stretchr/testify v1.8.4
 	golang.org/x/crypto v0.11.0
 	golang.org/x/exp v0.0.0-20230801115018-d63ba01acd4b
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20230807174057-1744710a1577
+	google.golang.org/grpc v1.57.0
```
ItalyPaleAle, line 16: Please use 1.56.2 which is the version currently used in the runtime, or we'll force the runtime to upgrade grpc late in the game: https://github.com/dapr/dapr/blob/ad5618711830510617e1b8fb2d0c6758f14ff6b9/go.mod#L67C3-L67C3

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	return f
+}
+
+func WithMetadata(md map[string]string) ErrorOption {
+	f := func(de *DaprError) {
+		de.metadata = md
+	}
+	return f
+}
```
JoshVanL, line 162: What's the use case for exposing these funcs? I don't think we should let these fields be mutable after creation.

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+	de := &DaprError{}
+	if errors.As(err, &de) {
+		if st, ese := de.newStatusError(); ese == nil {
+			return st, nil
+		}
+		return nil, err
+	}
+
+	return nil, err
+}
```
JoshVanL, line 180:
```suggestion
func (e *Error) ToGRPC() *status.Status {
```

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+		return true
+	}
+	return false
+}
+
+// NewDaprError create a new DaprError using the supplied metadata and ErrorOptions
+// **Note**: As this code is in `Feature Preview`, it will only continue processing
+// if the ErrorCodes is enabled
+func NewDaprError(err error, metadata map[string]string, options ...ErrorOption) *DaprError {
+	// The following condition can be removed once the
```
ItalyPaleAle, line 82: Please add a "TODO" comment here so we can track it

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+	de := &DaprError{
+		err:            err,
+		description:    err.Error(),
+		reason:         NoReason,
+		httpCode:       503,
+		grpcStatusCode: codes.Internal,
+	}
+
+	// Now apply any requested options
+	// to overr
```
ItalyPaleAle, line 102:
```suggestion
	// to override
```

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+	for _, option := range options {
+		option(de)
+	}
+
+	return de
+}
+
+// Error implements the error interface.
+func (c *DaprError) Error() string {
+	if c.err != nil {
```
ItalyPaleAle, line 112:
```suggestion
	if c != nil && c.err != nil {
```

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	if err == nil {
+		return nil
+	}
+
+	// Use default values
+	de := &DaprError{
+		err:            err,
+		description:    err.Error(),
```
ItalyPaleAle, line 95: Due to the check done below in `func (c *DaprError) Description() string `, we don't need this line and we can save a few bytes :)
```suggestion
```

errorcodes/errorcodes.go
JoshVanL: So as I understand- we are using this key to change behavior here based on a feature gate in dapr/dapr? I think it would be best for this package to be as simple as possible, and use this package or not based on the presence of the feature gate in dapr/dapr rather than needing to learn the feature gate here as well.

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+	return f
+}
+
+// *** GRPC Methods ***
+
+// FromDaprErrorToGRPC transforms the supplied error into
+// a GRPC Status. It assumes if the supplied error
+// is of type DaprError.
+// Otherwise, returns the original error.
+func FromDaprErrorToGRPC(err error) (*status.Status, error) {
```
ItalyPaleAle, line 170: We can simplify the code and not require a method to return a gRPC status.

If you change DaprError to implement `GRPCStatus()` then it is also used by the gRPC library. See an exmaple in APIError: https://github.com/dapr/dapr/blob/ad5618711830510617e1b8fb2d0c6758f14ff6b9/pkg/messages/api_error.go#L79-L83

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+		messages = append(messages, newResourceInfo(c.resourceInfo, c.err))
+	}
+
+	ste, stErr := status.New(c.grpcStatusCode, c.description).WithDetails(messages...)
+	if stErr != nil {
+		return nil, stErr
+	}
+
+	return ste, nil
+}
```
JoshVanL, line 197: Can this be pre-computed when we initialize the Error type in `New` so that we don't throw an error down stream when consumers want to get the `*status.Status`?

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+}
+
+// *** HTTP Methods ***
+
+// FromDaprErrorToHTTP transforms the supplied error into
+// a GRPC Status and then Marshals it to JSON.
+// It assumes if the supplied error is of type DaprError.
+// Otherwise, returns the original error.
+func FromDaprErrorToHTTP(err error) (int, []byte, error) {
```
ItalyPaleAle, line 224: Let's work with the runtime to make it so DaprError can be returned direclty by one of the errorResponders method in pkg/http

We already have half of the work done with the errorResponseValue interface: https://github.dev/dapr/dapr/blob/ad5618711830510617e1b8fb2d0c6758f14ff6b9/pkg/http/errors.go#L20-L23

This way we can pass DaprError objects to  `respondWithError` and it will just work: https://github.dev/dapr/dapr/blob/ad5618711830510617e1b8fb2d0c6758f14ff6b9/pkg/http/responses.go#L134-L135

ItalyPaleAle (changes requested):

errorcodes/errorcodes.go
JoshVanL, line 224:
```suggestion
func (e *Errro) ToHTTP(err error) (int, []byte) {
```

errorcodes/errorcodes.go
```diff
@@ -0,0 +1,241 @@
...
+	de := &DaprError{}
+	if errors.As(err, &de) {
+		if st, ese := de.newStatusError(); ese == nil {
+			resp, sej := statusErrorJSON(st)
+			if sej != nil {
+				return 0, nil, err
+			}
+
+			return de.httpCode, resp, nil
+		}
```
JoshVanL, line 234: Same, re: pre-computing values which may throw an error at initialization.

JoshVanL (changes requested):
Some comments from me!

errorcodes/errorcodes.go
robertojrojas: The `Set*` functions have been removed. The `With*` are used as optional at the time of creation.

errorcodes/errorcodes.go
artursouza: The feature flag is here because contrib also needs to know if the feature is enabled. This is not just runtime.

errors/errors.go
```diff
@@ -0,0 +1,220 @@
...
+
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/runtime/protoiface"
+)
+
+const (
+	ErrorCodesFeatureMetadataKey = "error_codes_feature"
```
artursouza, line 28: This key is for internal use only as a way to pass flags between runtime and contrib. It should not need to be exported and be removed if contrib ends up not needing it.

errorcodes/errorcodes.go
JoshVanL: I think it makes more sense for all the calling code to be feature gate aware rather than kit. Making kit aware of feature flags means that its functionality is tied to the version of runtime, rather than a separate self contained utility library. It also means there would now be a bi-directional relationship between kit and other repos, where it is only one way currently.

Contrib is already tightly coupled with the runtime version, what would be the implication of having the feature gate aware code moved there instead? Another option is to have a `kit/errors.Builder` struct which is given an `Options`, however I would assume this would end up being more verbose than the caller handling the feature gate logic.

errors/errors.go
```diff
@@ -0,0 +1,220 @@
...
+// New create a new Error using the supplied metadata and ErrorOptions
+// **Note**: As this code is in `Feature Preview`, it will only continue processing
+// if the ErrorCodes is enabled
+// TODO: @robertojrojas update when feature is ready.
+func New(err error, metadata map[string]string, options ...Option) *Error {
+	// The following condition can be removed once the
+	// Error Codes Feature is GA
+	if !featureEnabled(metadata) {
+		return nil
+	}
```
JoshVanL, line 82: Should this no be returning the `err` as is? This seems like a code smell that the caller is going to have to learn the feature gate anyway.

errorcodes/errorcodes.go
JoshVanL: @robertojrojas this comment is still relevant

errors/errors.go
```diff
@@ -0,0 +1,220 @@
...
+	return e.err.Error()
+}
+
+func WithErrorReason(reason string, httpCode int, grpcStatusCode codes.Code) Option {
+	f := func(er *Error) {
+		er.reason = reason
+		er.grpcStatusCode = grpcStatusCode
+		er.httpCode = httpCode
+	}
+	return f
```
JoshVanL, line 132: nit: no need to assign a variable which gets returned, and we should call the error variable `err` as this is consistent in Go. Same applies to for the other funcs below.

```suggestion
	return func(err *Error) {
		err.reason = reason
		err.grpcStatusCode = grpcStatusCode
		err.httpCode = httpCode
	}
```

errors/errors.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+// *** HTTP Methods ***
+
+// ToHTTP transforms the supplied error into
+// a GRPC Status and then Marshals it to JSON.
+// It assumes if the supplied error is of type Error.
+// Otherwise, returns the original error.
+func (e *Error) ToHTTP() (int, []byte) {
+	if resp, sej := protojson.Marshal(e.GRPCStatus().Proto()); sej == nil {
```
JoshVanL, line 202: What is sej?

errors/errors.go
```diff
@@ -0,0 +1,220 @@
...
+		metadata[ErrorCodesFeatureMetadataKey] = "true"
+	}
+}
+
+func featureEnabled(metadata map[string]string) bool {
+	_, ok := metadata[ErrorCodesFeatureMetadataKey]
+	return ok
+}
+
+// New create a new Error using the supplied metadata and ErrorOptions
```
JoshVanL, line 73:
```suggestion
// New create a new Error using the supplied metadata and Options
```

errors/errors.go
```diff
@@ -0,0 +1,220 @@
...
+
+type ResourceInfo struct {
+	Type string
+	Name string
+}
+
+// call this function to apply option
+type Option func(*Error)
+
+type Error struct {
```
JoshVanL, line 49: Please add comments explaining what these types are. As I understand it is a structured error type which helps to send the correct status codes and context over gRPC and HTTP, but that's not clear unless you read the whole file.

JoshVanL (changes requested):

errors/errors.go
robertojrojas: I was using an `err` variable and I used the name to differentiate. I'm renaming to `err`.

errorcodes/errorcodes.go
robertojrojas: Agreed! I'm removing the logic of the feature gate and having it on dapr/dapr + contrib.

errors/errors.go
robertojrojas: I've decided to move this key outside of `Kit` just to keep it clean as suggested by @JoshVanL

errors/errors.go
```diff
@@ -123,29 +111,35 @@ func (e *Error) Description() string {
...
 	}
-	return f
 }
 
+// WithResourceInfo used to pass ResourceInfo to the Error struct.
 func WithResourceInfo(resourceInfo *ResourceInfo) Option {
 	f := func(e *Error) {
 		e.resourceInfo = resourceInfo
 	}
 	return f
```
JoshVanL, line 129:
```suggestion
func WithResourceInfo(resourceInfo *ResourceInfo) Option {
	return func(e *Error) {
		e.resourceInfo = resourceInfo
	}
```

JoshVanL (changes requested):
Looks good- last couple nits and a question from me

artursouza (approved):

mukundansundar (approved):

JoshVanL (dismissed):

errors/errors.go
```diff
@@ -0,0 +1,212 @@
...
+		newErrorInfo(e.reason, e.metadata),
+	}
+
+	if e.resourceInfo != nil {
+		messages = append(messages, newResourceInfo(e.resourceInfo, e.err))
+	}
+
+	ste, stErr := status.New(e.grpcStatusCode, e.description).WithDetails(messages...)
+	if stErr != nil {
+		return status.New(e.grpcStatusCode, e.description)
```
ItalyPaleAle, line 176: Shouldn't this return an Internal Server Error? If I understand correctly, `WithDetails` returns an error if it fails to parse messages, which should indicate a development-time issue and not a runtime one.


```suggestion
		return status.New(codes.Internal, fmt.Sprintf("failed to create gRPC status message: %v", stErr))
```

errors/errors.go
```diff
@@ -0,0 +1,212 @@
...
+
+// HTTPCode returns the value of the HTTPCode property.
+func (e *Error) HTTPCode() int {
+	if e.httpCode == 0 {
+		return http.StatusInternalServerError
+	}
+	return e.httpCode
+}
+
+// JSONErrorValue implements the errorResponseValue interface.
```
ItalyPaleAle, line 205: Nit, for readers.

```suggestion
// JSONErrorValue implements the errorResponseValue interface (used by `github.com/dapr/dapr/pkg/http`).
```

errors/errors.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return http.StatusInternalServerError, []byte(err.Error())
+	}
+
+	return e.httpCode, resp
+}
+
+// HTTPCode returns the value of the HTTPCode property.
+func (e *Error) HTTPCode() int {
+	if e.httpCode == 0 {
```
ItalyPaleAle, line 199: Nil-check
```suggestion
    if e == nil {
        return http.StatusOK
    }
	if e.httpCode == 0 {
```

errors/errors.go
```diff
@@ -0,0 +1,212 @@
...
+		return http.StatusInternalServerError
+	}
+	return e.httpCode
+}
+
+// JSONErrorValue implements the errorResponseValue interface.
+func (e *Error) JSONErrorValue() []byte {
+	b, err := protojson.Marshal(e.GRPCStatus().Proto())
+	if err != nil {
+		return []byte(err.Error())
```
ItalyPaleAle, line 209: This is not going to work because `err.Error()` is not a JSON-encoded string.

```suggestion
        errJSON, _ := json.Marshal(fmt.Sprintf("failed to encode proto to JSON: %v", err))
		return errJSON
```

errors/errors.go
```diff
@@ -0,0 +1,212 @@
...
+// *** HTTP Methods ***
+
+// ToHTTP transforms the supplied error into
+// a GRPC Status and then Marshals it to JSON.
+// It assumes if the supplied error is of type Error.
+// Otherwise, returns the original error.
+func (e *Error) ToHTTP() (int, []byte) {
+	resp, err := protojson.Marshal(e.GRPCStatus().Proto())
+	if err != nil {
+		return http.StatusInternalServerError, []byte(err.Error())
```
ItalyPaleAle, line 191: Same, if the expecation is that the response is JSON, the value must be JSON-encoded.

```suggestion
        errJSON, _ := json.Marshal(fmt.Sprintf("failed to encode proto to JSON: %v", err))
		return http.StatusInternalServerError, errJSON
```

errors/errors.go
```diff
@@ -0,0 +1,212 @@
...
+func (e *Error) GRPCStatus() *status.Status {
+	messages := []protoiface.MessageV1{
+		newErrorInfo(e.reason, e.metadata),
+	}
+
+	if e.resourceInfo != nil {
+		messages = append(messages, newResourceInfo(e.resourceInfo, e.err))
+	}
+
+	ste, stErr := status.New(e.grpcStatusCode, e.description).WithDetails(messages...)
```
ItalyPaleAle, line 174: Per the docs, `protoiface` should not be imported directly. There's an alias in the `github.com/golang/protobuf/proto` package, but that package is deprecated.

We can get around all of that (and avoid allocating a slice too, in the process) with something like:
```suggestion
	var stErr error
	ste := status.New(e.grpcStatusCode, e.description)
	if e.resourceInfo != nil {
		ste, stErr = ste.WithDetails(newErrorInfo(e.reason, e.metadata), newResourceInfo(e.resourceInfo, e.err))
	} else {
		ste, stErr = ste.WithDetails(newErrorInfo(e.reason, e.metadata))
	}
```

ItalyPaleAle:
Thanks for your patience @robertojrojas . It looks much better, and I have only some relatively minor comments (the most critical ones are about making sure that the methods that are supposed to return JSON, always return valid JSON).

On an unrelated note, because this PR changes the `go.mod` file, we should not merge until the merge window for 1.13 opens in dapr/kit. Otherwise, importing this package in dapr/dapr for 1.12 too will cause the dependency tree to change, and it's too late in the endgame for that.

errors/errors.go
```diff
@@ -0,0 +1,300 @@
...
+		errJSON, _ := json.Marshal(fmt.Sprintf("failed to encode proto to JSON: %v", err))
+		return errJSON
+	}
+	return b
+}
+
+// HTTPStatusFromCode converts a gRPC error code into the corresponding HTTP response status.
+// https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/errors.go#L15
+// See: https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
+func HTTPStatusFromCode(code codes.Code) int {
```
ItalyPaleAle, line 225: Please move this and the method below to a separate package (maybe `dapr/kit/grpccodes`? naming is hard - you pick)

We can then update dapr/dapr to use that package, but should be on its own

errors/errors.go
```diff
@@ -0,0 +1,221 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/dapr/kit/grpccodes"
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/encoding/protojson"
```
ItalyPaleAle, line 25:
```suggestion
	"google.golang.org/genproto/googleapis/rpc/errdetails"
	"google.golang.org/grpc/codes"
	"google.golang.org/grpc/status"
	"google.golang.org/protobuf/encoding/protojson"

	"github.com/dapr/kit/grpccodes"
```

ItalyPaleAle (approved):

grpccodes/grpccodes.go
```diff
@@ -0,0 +1,100 @@
...
+		return http.StatusNotImplemented
+	case codes.Internal:
+		return http.StatusInternalServerError
+	case codes.Unavailable:
+		return http.StatusServiceUnavailable
+	case codes.DataLoss:
+		return http.StatusInternalServerError
+	}
+
+	return http.StatusInternalServerError
```
JoshVanL, line 64: nit:

```suggestion
        default:
	  return http.StatusInternalServerError
	}
```

JoshVanL (changes requested):
Couple more nits from me but otherwise LGTM

JoshVanL (approved):

ItalyPaleAle (approved):

PR #59, opened 2023-08-11
Always wait in close, Speed up eventqueue tests & don't pass clock

JoshVanL:
Adds test for `Close()`.

Always wait for the processor to return go routines when calling `Close()`.

Speeds up the eventqueue tests by ~`4x` (`13` -> `3` seconds) by relying less on the wall clock.

Don't require passing a clock to the processor as use a real clock by default. A custom clock can be given through `WithClock` whose function is available when the `unit` build tag is added.

/cc @ItalyPaleAle

### DIFF
--- a/events/queue/eventqueue_test.go
+++ b/events/queue/eventqueue_test.go
@@ -16,8 +16,6 @@ package queue
 import (
 	"fmt"
 	"time"
-
-	kclock "k8s.io/utils/clock"
 )
 
 // queueableItem is an item that can be queued and it's used for testing.
@@ -38,9 +36,6 @@ func (r queueableItem) ScheduledTime() time.Time {
 }
 
 func ExampleProcessor() {
-	// Init a clock using k8s.io/utils/clock
-	clock := kclock.RealClock{}
-
 	// Method invoked when an item is to be executed
 	executed := make(chan string, 3)
 	executeFn := func(r *queueableItem) {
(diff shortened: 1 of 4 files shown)


eventqueue/processor_unit.go
```diff
@@ -0,0 +1,27 @@
...
+*/
+
+package eventqueue
+
+import (
+	kclock "k8s.io/utils/clock"
+)
+
+// WithClock sets the clock used by the processor. Used for testing.
+func (p *Processor[T]) WithClock(clock kclock.Clock) *Processor[T] {
```
ItalyPaleAle, line 24: I like this, good idea

eventqueue/processor_test.go
```diff
@@ -53,31 +52,14 @@ func TestProcessor(t *testing.T) {
...
-		// The signal is sent in a background goroutine, so we need to use a wall clock here
-		runtime.Gosched()
 		select {
 		case r := <-executeCh:
 			t.Fatalf("received unexpected item: %s", r.Name)
-		case <-time.After(500 * time.Millisecond):
+		case <-time.After(100 * time.Millisecond):
 			// all good
 		}
 	}
```
ItalyPaleAle, line 64: The reason for using 500ms, as well as `runtime.Gosched()` (which could be kept anyways? it shouldn't cause issues) is that our Actions runner as SLOW... I am concerned that having a timeout that's too low could cause false positives.

eventqueue/processor_test.go
```diff
@@ -118,9 +100,11 @@ func TestProcessor(t *testing.T) {
 			require.NoError(t, err)
 		}
 
-		// Advance tickers by 1500ms to trigger the first item
+		assert.Eventually(t, clock.HasWaiters, time.Second, 100*time.Millisecond)
+
+		// Advance tickers by 1s to trigger the first item
 		t.Log("Waiting for signal 1")
-		advanceTickers(500*time.Millisecond, 3)
```
ItalyPaleAle, line 123: I made it advance in steps before also to assert that things do need 3 ticks to trigger. That was my thought at least.

eventqueue/processor_test.go
```diff
@@ -164,17 +149,21 @@ func TestProcessor(t *testing.T) {
 		err = processor.Dequeue("4")
 		require.NoError(t, err)
 
+		assert.Equal(t, 3, processor.queue.Len())
+
 		// Advance tickers and assert messages are coming in order
 		for i := 1; i <= 5; i++ {
+			assert.Eventually(t, clock.HasWaiters, time.Second, 100*time.Millisecond)
```
ItalyPaleAle, line 156: Should we use a `require` here?

eventqueue/processor_test.go
```diff
@@ -139,23 +123,24 @@ func TestProcessor(t *testing.T) {
...
-			advanceTickers(time.Second, 1)
 			received := assertExecutedItem(t)
 			assert.Equal(t, expect, received.Name)
+			clock.Step(time.Second)
 		}
 	})
 
 	t.Run("dequeue item", func(t *testing.T) {
+		assert.Equal(t, 0, processor.queue.Len())
+		assert.False(t, clock.HasWaiters())
```
ItalyPaleAle, line 134: Should we use a `require` here?

eventqueue/processor.go
```diff
@@ -93,17 +94,15 @@ func (p *Processor[T]) Dequeue(key string) error {
 // Close stops the processor.
 // This method blocks until the processor loop returns.
 func (p *Processor[T]) Close() error {
-	if !p.stopped.CompareAndSwap(false, true) {
-		// Already stopped
+	defer p.wg.Wait()
```
ItalyPaleAle, line 97: Ok I understand now what you were saying a couple of weeks back. Ok, this makes sense.

artursouza (dismissed):

artursouza:
Please, resolve conflicts.

eventqueue/processor_test.go
JoshVanL: Fair enough, I'll revert it back.

eventqueue/processor_test.go
JoshVanL: What is the thinking behind needing 3 distinct ticks, rather than a single larger tick?

JoshVanL:
@artursouza done.

artursouza (approved):

artursouza (approved):

PR #64, opened 2023-10-21
Adds events/ratelimiting/coalecing

JoshVanL:
events/ratelimiting is a new utility package to provides helpers which can be used to rate limit events.

Coalecing is a new helper that will exponentially rate limit events. It will coalesce events into a single event if they occur within the same rate limiting window. Coalesce also has the option to forcibly fire an event when the number of events reaches a certain threshold. Added to prevent events from never being fired in a high throughput scenario.


events/ratelimiting/ratelimiter_test.go
```diff
@@ -0,0 +1,21 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ratelimiting
+
+import "testing"
+
+func TestRateLimiters(*testing.T) {
+	var _ RateLimiter = new(coalescing)
```
ItalyPaleAle, line 19: Change to a compile-time assertion in `ratelimiter.go`

```go
var _ RateLimiter = (*coalescing)(nil)
// In unit.go
var _ RateLimiterWithTicker = (*coalescing)(nil)
```

(These don't cause allocations)

events/ratelimiting/coalescing.go
```diff
@@ -0,0 +1,256 @@
...
+	closed  atomic.Bool
+}
+
+func NewCoalescing(opts OptionsCoalescing) (RateLimiter, error) {
+	initialDelay := time.Millisecond * 500
+	if opts.InitialDelay != nil {
+		initialDelay = *opts.InitialDelay
+	}
+	if initialDelay < 0 {
+		return nil, errors.New("initial delay must be > 0")
```
ItalyPaleAle, line 73: You say "must be > 0" but the check allows everything ">= 0"

events/ratelimiting/coalescing.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	maxDelay := time.Second * 5
+	if opts.MaxDelay != nil {
+		maxDelay = *opts.MaxDelay
+	}
+	if maxDelay < 0 {
+		return nil, errors.New("max delay must be > 0")
```
ItalyPaleAle, line 81: Ditto

events/ratelimiting/coalescing.go
```diff
@@ -0,0 +1,256 @@
...
+
+		if !c.timer.Stop() {
+			<-c.timer.C()
+		}
+
+		// Setup backoff. Backoff is exponential. If initial is 500ms and max is
+		// 5s, the backoff will follow:
+		// 500ms, 1s, 2s, 4s, 5s, 5s, 5s, ...
+		if c.currentDur < c.maxDelay {
+			c.backoffFactor *= 2
```
ItalyPaleAle, line 182: Do we need some jitter?

events/ratelimiting/coalescing.go
```diff
@@ -0,0 +1,256 @@
...
+	c.lock.Unlock()
+	defer c.wg.Done()
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	for {
+		// If the timer doesn't exist yet, we're waiting for the first event (which
+		// will fire immediately when received).
+		if !c.hasTimer.Load() {
```
ItalyPaleAle, line 123: I think this could be simplified to keep it DRY (sing ereading from a nil channel is allowed and it never receives a message)

```go
        var timerCh <-chan Time
		if c.hasTimer.Load() {
			timerCh = c.timer.C()
		}

		select {
		case <-ctx.Done():
			return nil
		case <-c.closeCh:
			cancel()
			return nil

		case <-c.inputCh:
			c.handleInputCh(ctx, ch)
		case <-timerCh:
			c.handleTimerFired(ctx, ch)
		}
```

events/ratelimiting/ratelimiter_test.go
JoshVanL: Question: What is the benefit of having this in the compiled file? The benefit of having it in the test file is that it doesn't clutter the implementation code.

events/ratelimiting/coalescing.go
JoshVanL: Fixed

events/ratelimiting/coalescing.go
JoshVanL: Fixed

events/ratelimiting/coalescing.go
JoshVanL: I initially did have some jitter but I don't think it makes sense in this scenario since we are not concerned with contention as we are coalescing the requests and not responding to clients which would all re-request at the same time on backoff.

events/ratelimiting/coalescing.go
JoshVanL: Thanks, this is a lot cleaner :+1:

events/ratelimiting/ratelimiter_test.go
ItalyPaleAle: I think it's just the pattern I've generally seen. It makes errors immediately visible by the compiler (it is also something that should have no or negligible impact on the compiled code) 🤷‍♂️

Also, would fix the errors the linter is reporting :)

ItalyPaleAle (commented):
Please fix the linter issues

events/ratelimiting/ratelimiter_test.go
JoshVanL: The linter is erroring because we are using a GitHub native action to run the linter rather than the Makefile, so the `unit` build tag is not being applied. I have opened https://github.com/dapr/kit/pull/67

events/ratelimiting/unit.go
```diff
@@ -0,0 +1,32 @@
...
+type RateLimiterWithTicker interface {
+	RateLimiter
+	WithTicker(clock.WithTicker)
+}
+
+func (c *coalescing) WithTicker(clock clock.WithTicker) {
+	c.clock = clock
+}
+
+var _ RateLimiterWithTicker = new(coalescing)
```
ItalyPaleAle, line 32: Using `new` causes the allocation of the object which stays around forever

```suggestion
var _ RateLimiterWithTicker = (*coalescing)(nil)
```

events/ratelimiting/coalescing.go
```diff
@@ -0,0 +1,251 @@
...
+		c.lock.Lock()
+		c.wg.Wait()
+		c.lock.Unlock()
+	}()
+	if c.closed.CompareAndSwap(false, true) {
+		close(c.closeCh)
+	}
+}
+
+var _ RateLimiter = new(coalescing)
```
ItalyPaleAle, line 251: Using `new` causes the allocation of the object which stays around forever

```suggestion
var _ RateLimiter = (*coalescing)(nil)
```

ItalyPaleAle (changes requested):

events/ratelimiting/unit.go
JoshVanL: Done

events/ratelimiting/coalescing.go
JoshVanL: Done

Makefile
```diff
@@ -77,7 +77,7 @@ test-race:
 .PHONY: lint
 lint:
 	# Due to https://github.com/golangci/golangci-lint/issues/580, we need to add --fix for windows
-	$(GOLANGCI_LINT) run --timeout=20m
+	$(GOLANGCI_LINT) run --timeout=20m --build-tags unit
```
ItalyPaleAle, line 80:
```suggestion
	$(GOLANGCI_LINT) run --timeout=20m
```

ItalyPaleAle (approved):

PR #65, opened 2023-10-21
Fixing running tests on aarch64 arm machines

JoshVanL:
Also adds `-race` flag to test, and enable `unit` build tag in tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,10 @@ else ifeq ($(shell echo $(LOCAL_ARCH) | head -c 5),armv8)
 	TARGET_ARCH_LOCAL=arm64
 else ifeq ($(shell echo $(LOCAL_ARCH) | head -c 4),armv)
 	TARGET_ARCH_LOCAL=arm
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -61,7 +65,7 @@ endif
 ################################################################################
 .PHONY: test
 test:
-	go test ./... $(COVERAGE_OPTS) $(BUILDMODE)
+	CGO_ENABLED=1 go test -race -tags unit ./... $(COVERAGE_OPTS) $(BUILDMODE)
```
ItalyPaleAle, line 68: Let’s have a separate command for this. This makes test run a lot slower. It is also using CGo which we don’t use to build the final binary.

ItalyPaleAle (changes requested):

Makefile
JoshVanL: Done

ItalyPaleAle (approved):

PR #66, opened 2023-10-22
Use mocked clock in cron tests, rather than wall clock

JoshVanL:
PR updates the cron tests to use a mocked clock instead of the wall in order to speed up the tests from ~40s to ~1.5s. This also has the benefit of controlling time so that tests are more deterministic.

### DIFF
--- a/cron/chain_test.go
+++ b/cron/chain_test.go
@@ -23,6 +23,9 @@ import (
 	"sync"
 	"testing"
 	"time"
+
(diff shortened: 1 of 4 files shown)


ItalyPaleAle (approved):

ItalyPaleAle:
Please fix DCO

JoshVanL:
@ItalyPaleAle Done

PR #67, opened 2023-10-23
Adds `--build-tags=unit` to golang lint config file

JoshVanL:


.github/workflows/kit.yml
```diff
@@ -55,6 +55,7 @@ jobs:
         with:
           version: ${{ env.GOLANGCI_LINT_VER }}
           skip-cache: true
+          args: --build-tags=unit
```
ItalyPaleAle, line 58: Please modify the `.golangci-lint.yaml` file instead, so it runs when developers run linter locally too

JoshVanL: Yes, good shout! Done

ItalyPaleAle (approved):

PR #77, opened 2023-11-30
Feat standardize err msgs

cicoyle:
# Description

Standardizing error codes thorough out Dapr with this errors pkg. 

NOTE: I added the gRPC err_detail type to show up in the json output in the same manner as `@type` since that is how gRPC does it and to not conflict with the field `type` for `PreconditionFailure_Violation`.  Also, putting the `@` sign puts the type as the first field in the field output ordering.

I will update the README example once I finalize my dapr runtime code in a followup PR.

image
image

## Issue reference

https://github.com/dapr/proposals/blob/main/0009-BCIRS-error-handling-codes.md

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


errors/errors.go
```diff
@@ -15,209 +15,310 @@ package errors
...
+	Tag string
 }
 
 // New create a new Error using the supplied metadata and Options
-func New(err error, metadata map[string]string, options ...Option) *Error {
-	if err == nil {
-		return nil
-	}
-
+func New(grpcCode grpcCodes.Code, httpCode int, message string, tag string) *Error {
```
ItalyPaleAle, line 54: Since `Error` implements the error interface, consider changing the return type to `error`:


```suggestion
func New(grpcCode grpcCodes.Code, httpCode int, message string, tag string) error {
```

errors/errors.go
```diff
@@ -15,209 +15,310 @@ package errors
...
 	return ""
 }
 
-// Unwrap implements the error unwrapping interface.
-func (e *Error) Unwrap() error {
-	if e == nil {
-		return nil
-	}
-	return e.err
+// String returns the string representation, useful for debugging.
```
ItalyPaleAle, line 75: This is now used in `Error()` so not just for debugging

errors/errors.go
```diff
@@ -15,209 +15,310 @@ package errors
...
 
-func newErrorInfo(reason string, md map[string]string) *errdetails.ErrorInfo {
-	return &errdetails.ErrorInfo{
-		Domain:   errorInfoDefaultDomain,
-		Reason:   reason,
-		Metadata: md,
-	}
+// WithVars returns a copy of the error with the message going through fmt.Sprintf with the arguments passed to this method.
+func (e *Error) WithVars(a ...any) *Error {
+	eCopy := *e
```
ItalyPaleAle, line 108: Not 100% but I don't think this copies the entire object. At least the slice that is contained (`Details`) may not be copied

errors/errors.go
```diff
@@ -15,209 +15,310 @@ package errors
...
+	// This will get overwritten later if there is an ErrorInfo code
+	httpStatus := e.Tag
+	if httpStatus == "" {
+		httpStatus = http.StatusText(e.HttpCode)
 	}
 
-	return e.httpCode
-}
+	// Change output to match prior output to not break users
+	errJson := map[string]interface{}{
```
ItalyPaleAle, line 164: Could you use a strongly-typed struct here rather than a map?

errors/errors.go
```diff
@@ -15,209 +15,310 @@ package errors
...
+			case *errdetails.Help_Link:
+				desc := typedDetail.ProtoReflect().Descriptor()
+				detailMap := map[string]interface{}{
+					"@type":       typeGoogleAPI + desc.FullName(),
+					"description": typedDetail.Description,
+					"url":         typedDetail.Url,
+				}
+				errJson["details"].([]interface{})[i] = detailMap
+			default:
+				log.Debugf("Failed to convert error details due to incorrect type. \nSee types here: https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto. \nDetail: %s", detail)
```
ItalyPaleAle, line 295: Perhaps this should panic as it would indicate a development-time error only?

cicoyle: I agree we should catch this in development time, but Im worried that not all http cases get tested and this would come up during http only, not grpc. This is an edge case we shouldn't really get into, but just in case I feel like it's better to still give the details and not panic.

errors/errors.go
cicoyle: I didn't want to go the struct route and have all the diff possible types defined bc there are so many err_details, and since I didn't want to do that for the err_details to be consistent I went with this approach.

ItalyPaleAle: Fine to use a map for details, but `errJSON` itself seems to have just a handful of keys: errorCode (string? I think), message (string), details (map[string]any)

errors/errors.go
cicoyle: this func will be removed entirely

errors/errors.go
cicoyle: Im not sure I'm 100% following what you are suggesting. The point of standardizing the errors in this manner was to have a new error with all the status and details additions, if I switch to returning a standard lib error, then we lose all that additional error info. If you were suggesting making my `Error` struct `error` the IDE complains about the name, as that is the standard lib error.

ItalyPaleAle: `error` is an interface, so you can return `*Error` that implements `error`.

In the caller, then, they can use `errors.As` to check if an `error` is of type `*Error`.

See: https://earthly.dev/blog/golang-errors/

cicoyle: I think I see what you are saying. What are your thoughts on this: Here we leave the `New()` func to return the `*Error`, then inside Dapr, when calling this New func, It'll look like (note its returning the standard `error`:
```
func PubSubNotFound(name string, pubsubType string, metadata map[string]string) error {
	message := fmt.Sprintf("pubsub %s is not found", name)

	return kitErrors.New(
		grpcCodes.NotFound,
		http.StatusBadRequest,
		message,
		kitErrors.CodePrefixPubSub+kitErrors.CodeNotFound,
	).
		WithErrorInfo(kitErrors.CodePrefixPubSub+kitErrors.CodeNotFound, metadata).
		WithResourceInfo(pubsubType, name, "", message)
}
```

Where upon seeing this err inside dapr code, say `http/api.go` in my case, I can do as you mentioned ^ where I have a conditional checking the type like the following:

```
		var kitError *kitErrors.Error
		if errors.As(validationErr, &kitError) {
		...
		}
```
This way in dapr we see the `error` type, and can check the errType with the `errors.As` as you mentioned. 

Upon making these changes, I still see the extra err_details 👍🏻

JoshVanL: I am also +1 on returning the primitive `error` type on `New` if we can. This is to follow the Go language decision to treat all errors the "same" until a type check is relevant. This is to make the return `err` variable more portable given a shared base type. You can see this done in the stdlib where, for example, all `net/url` functions will return a `net/url.Error` as a primitive `error` even though it will probably be of the `Error` type.

Do we expect it to be the case that the kit `*Error` will want to be mutated after it has been initially constructed? I _think_ we don't.. which means it's equally important that the type is immutable by the caller. All fields should also then be made private, and values returned through exported methods (if they need to be exposed at all). To keep the "builder" style constructor which I think we want, we can create an intermediate `Builder` type that results in an immutable `error` type.

artursouza: In this case, it seems better to implement the Builder pattern and make Error immutable. It is a common design pattern when building an instance is "too complex".

elena-kolevska: Could we have a situation where we need to create an error deeper in the code, and then extend it with more details (like resource info for example) somewhere on the api layer? I ran into a situation like this with checking for invalid keyname in state stores. The method that checks the keyname doesn't know the name of the state store we're working with, for example, which means we can't construct the ResourceInfo error detail in there, but we can extend the error once it gets to the api level. There are ways to go around this, of course, if we want to keep errors immutable, but it's probably worth considering.

cicoyle: This PR has been updated accordingly: to include the builder pattern (see readme for example usage) and includes an `AddDetails` to the `Error` struct for adding details in runtime when needed.

errors/messages.go
```diff
@@ -0,0 +1,33 @@
...
+
+const (
+	ErrMsgDomain = "dapr.io"
+
+	errStringFormat = "api error: code = %s desc = %s"
+
+	typeGoogleAPI = "type.googleapis.com/"
+
+	// Messages
+	MsgStateGet        = "fail to get %s from state store %s: %s"
```
artursouza, line 24: nit: why is this one in the present tense?

artursouza (changes requested):

artursouza:
LGTM, just a small change.

artursouza (dismissed):

errors/messages.go
cicoyle: @elena-kolevska can you confirm? I think it may be due to keeping the legacy messages that currently exist in dapr

errors/errors.go
```diff
@@ -15,209 +15,389 @@ package errors
...
-const (
-	resourceInfoDefaultOwner = "dapr-components"
-	errorInfoDefaultDomain   = "dapr.io"
-	errorInfoResonUnknown    = "UNKNOWN_REASON"
-)
+var log = logger.NewLogger("dapr.kit")
+
+// Error implements the Error interface and the interface that complies with "google.golang.org/grpc/status".FromError().
+// It can be used to send errors to HTTP and gRPC servers, indicating the correct status code for each.
+type Error struct {
```
ItalyPaleAle, line 35: `errors.Error` is a bit of a stutter? @JoshVanL thoughts?

errors/messages.go
```diff
@@ -0,0 +1,33 @@
...
+
+const (
+	ErrMsgDomain = "dapr.io"
+
+	errStringFormat = "api error: code = %s desc = %s"
+
+	typeGoogleAPI = "type.googleapis.com/"
+
+	// Messages
+	MsgStateGet        = "failed to get %s from state store %s: %s"
```
ItalyPaleAle, line 24: Do these constants need to be exported? Since they are formatting consts, should they be private?

errors/errors.go
```diff
@@ -15,209 +15,389 @@ package errors
...
+}
+
+/*** GRPC Methods ***/
+
+// GRPCStatus returns the gRPC status.Status object.
+func (e Error) GRPCStatus() *status.Status {
+	stat := status.New(e.grpcCode, e.message)
+
+	// convert details from proto.Msg -> protoiface.MsgV1
+	var convertedDetails []protoiface.MessageV1
```
ItalyPaleAle, line 115:
```suggestion
	convertedDetails := make([]protoiface.MessageV1, 0, len(e.details))
```

errors/errors.go
```diff
@@ -15,209 +15,389 @@ package errors
...
-	resourceInfo   *ResourceInfo
-}
-
-// New create a new Error using the supplied metadata and Options
-func New(err error, metadata map[string]string, options ...Option) *Error {
-	if err == nil {
-		return nil
-	}
+// ErrorJSON is used to build the error for the HTTP Methods json output
+type ErrorJSON struct {
```
ItalyPaleAle, line 59: Does this need to be exported?

errors/errors.go
```diff
@@ -15,209 +15,389 @@ package errors
...
+	// Tag is a string identifying the error, used with HTTP responses only.
+	tag string
 }
 
-// Option allows passing additional information
-// to the Error struct.
-// See With* functions for further details.
-type Option func(*Error)
+// ErrorBuilder is used to build the error
+type ErrorBuilder struct {
```
ItalyPaleAle, line 54: Does this need to be exported?

errors/errors.go
```diff
@@ -15,209 +15,389 @@ package errors
...
-		e.description = description
+	// Handle err details
+	details := e.details
+	if len(details) > 0 {
+		errJSON.Details = make([]any, len(details))
+		for i, detail := range details {
+			// cast to interface to be able to do type switch
+			// over all possible error_details defined
+			// https://github.com/googleapis/go-genproto/blob/main/googleapis/rpc/errdetails/error_details.pb.go
+			switch typedDetail := detail.(type) {
```
ItalyPaleAle, line 163: This may be hard to maintain in the future. 2 non-binding, but strong recommendations:

1. Please write a test to ensure the completness of the cases here compared with all the possible values. This way, if the upstream library adds a new case, we won't miss it.
   - We do a pretty bad job when updating deps and generally just update them without thinking. This means if the upstream library adds a case, we are almost certainly going to miss that.
2. Consider moving this `switch` to a new function, possibly even in a new file, so it's easier to test this (also for completness) and maintain it in the future.

errors/errors.go
```diff
@@ -15,209 +15,389 @@ package errors
...
+// WithDetails is used to pass any error details to the Error struct.
+func (b *ErrorBuilder) WithDetails(details ...proto.Message) *ErrorBuilder {
+	b.err.details = append(b.err.details, details...)
+
 	return b
 }
+
+// Build builds our error
+func (b *ErrorBuilder) Build() error {
+	// Check for ErrorInfo, since its required per the proposal
```
ItalyPaleAle, line 388: Typo
```suggestion
	// Check for ErrorInfo, since it's required per the proposal
```

errors/messages.go
elena-kolevska: Yes! This is an existing message we moved from dapr/dapr. I think it shouldn't be a problem if we change an error message, as long as we keep the error codes the same. If you agree, I will update this to match the other error strings.

errors/errors.go
```diff
@@ -15,209 +15,389 @@ package errors
...
+	containsErrorInfo := false
+	for _, detail := range b.err.details {
+		if _, ok := detail.(*errdetails.ErrorInfo); ok {
+			containsErrorInfo = true
+			break
+		}
+	}
+
+	if !containsErrorInfo || len(b.err.details) == 0 {
```
ItalyPaleAle, line 397: Nit, remove redundant check
```suggestion
	if !containsErrorInfo {
```

ItalyPaleAle (changes requested):

errors/messages.go
cicoyle: we are using these in dapr

errors/errors.go
cicoyle: fair point

errors/messages.go
ItalyPaleAle: Ok please add a comment that these are used in Dapr.

However consider whether keeping them in dapr/kit is the right place. If you need to add or update constants, having them in a separate repo will make your life a lot harder

errors/errors.go
cicoyle: is my existing test not sufficient: `TestError_JSONErrorValue`. I add one, multiple, and each of these err_details

ItalyPaleAle: No, that's a different test case I'm talking about.

In this `switch` you are adding a `case` for each possible type of `error_details` that is defined in https://github.com/googleapis/go-genproto/blob/main/googleapis/rpc/errdetails/error_details.pb.go (as the comment says).

The problem is that this is a 3rd-party library we have no control over. If Google decides to add a new possible type, then we will be missing a `case` for that.

We need to add a test that checks to make sure we have a `case` defined for each possible `error_details` defined in the `github.com/googleapis/go-genproto/googleapis/rpc/errdetails` package, today and in the future.

cicoyle: oh I see what you are saying now. That is partly why I don't panic when you commented a while ago about panic-ing if a user adds an unknown err_detail... I still add the err_detail if it is unknown, but I can add an explicit test for ensuring the types as you suggested

errors/messages.go
cicoyle: Initially we thought we might use these in both dapr and components-contrib, but for now I'll rm them and add them back if we find the need to use the same messages in both dapr and components-contrib

errors/messages.go
cicoyle: I'll rm these

errors/messages.go
cicoyle: rm-ing them

errors/errors.go
cicoyle: I guess not, confirming

errors/errors.go
```diff
@@ -15,209 +15,389 @@ package errors
...
+		log.Errorf("Must include ErrorInfo in error details. Error: ", b.err)
+		panic("Must include ErrorInfo in error details.")
+	}
+
+	return b.err
+}
```
JoshVanL, line 403: Do we want to include `func FromError(err error) (*Error, bool)` and `func IsError(err error) bool` helper functions?

JoshVanL (changes requested):

errors/errors.go
cicoyle: I added `TestEnsureAllErrDetailsCovered`

errors/errors.go
cicoyle: I unexported it in my recent commit, but CI fails: 
```  
Warning: unexported-return: exported func NewBuilder returns unexported type *errors.errorBuilder, which can be annoying to use (revive)
```

So I will re-export it again

errors/errors.go
cicoyle: Elena has a PR to the go-sdk showing using the `status.Convert(err)`. I don't think we necessarily need to add the `FromError`. I do have a `func (e *Error) Is(targetI error) bool {` func defined in `errors.go`. Lmk if you think I need to include the funcs you mentioned above still.

errors/errors_test.go
```diff
@@ -810,3 +813,77 @@ func TestErrorBuilder_Build(t *testing.T) {
...
+	}
+
+	if packages.PrintErrors(pkgs) > 0 {
+		t.Errorf("ensure package is correct: %v", packages.ListError)
+	}
+
+	// This is hard-coded from the switch statement in errors.go to ensure we stay up
+	// to date on the error types we support, and to ensure we update our supported
+	// error types when google adds to their error details
+	mySwitchTypes := []string{
```
ItalyPaleAle, line 837: Is there a way this can not be hardcoded? Maybe using reflection or the ast package?

Another option could be not to hardcode this list. Assuming we don't care if we have extra cases in the `switch`:

1. Export the `switch` statement in its own function so it's easier to test
2. Iterate on all the types exported by the Google's package
3. Invoke the function
4. Confirm the function can handle that case

cicoyle: I now dynamically grab the values using the ast pkg. Its very ugly. I've not used the pkg before, so if you can make it cleaner I'm open to changing it, but it gets the job done dynamically and I added comments on the ast bits since its not pretty.

ItalyPaleAle: This looks pretty nice to me now @cicoyle . I've only used the ast pkg once so I'm no expert myself, but if it works... :) Thanks for creating this test, looks really good work

errors/errors.go
elena-kolevska: We can't have a `FromError` with that signature, because it doesn't contain enough information for our error; namely, the `ErrorInfo` status detail is required, which means we need a status code.

elena-kolevska: Ok, I guess Josh meant to use the FromErrors() as a shortcut for a type conversion, in that case, that makes sense to me.

elena-kolevska (approved):

errors/errors.go
cicoyle: I think I Implemented this per the above request. The usage of `FromError()` looks like:
```
	standardizedErr, ok := kitErrors.FromError(err)
	if ok {
		fasthttpRespond(reqCtx, fasthttpResponseWithKitError(standardizedErr, nil))
		return
	}
```

errors/errors.go
```diff
@@ -113,6 +113,20 @@ func (e *Error) AddDetails(details ...proto.Message) *Error {
 	return e
 }
 
+// FromError takes in an error and returns back the kitError if it's that type under the hood
+func FromError(err error) (*Error, bool) {
+	if err == nil {
+		return nil, false
+	}
+
+	kitErr, ok := err.(Error)
```
ItalyPaleAle, line 122: Use `errors.As` here, which supports the case where `Error` is wrapped inside another error.

errors/errors_test.go
```diff
@@ -887,3 +942,29 @@ func containsType(types []string, target string) bool {
...
 	return false
 }
+
+func TestFromError(t *testing.T) {
+	result, ok := FromError(nil)
+	if result != nil || ok {
+		t.Errorf("Expected result to be nil and ok to be false, got result: %v, ok: %t", result, ok)
+	}
+
+	kitErr := Error{
```
ItalyPaleAle, line 952: Please also add a test with `wrapped := fmt.Errorf("wrapped: %w", kitErr)`

errors/errors.go
cicoyle: good idea, done.

JoshVanL (approved):
Awesome work 😄

ItalyPaleAle (approved):

PR #82, opened 2024-01-14
Change `events/batcher` to use `events/queue` as backend.

JoshVanL:
The current implementation of `events/batcher` has issues with deadlocks when used by `fswatcher` which are causing unit tests to fail in dapr/dapr CI, and may even cause some issues in Dapr deployments.. PR updates `batcher` to use `queue` as the backend, removing these deadlocks.

PR also updates `events/queue` to make the key index type a generic comparable type, enabling it work on other comparable types like `int`, `bool`, `float64` etc. rather than just `string`. This is useful as there are already consumers of batcher indexing on `int`.

### DIFF
--- a/events/batcher/batcher.go
+++ b/events/batcher/batcher.go
@@ -27,10 +27,10 @@ import (
 // the added event channel subscribers. Events are sent to the channels after
 // the interval has elapsed. If events with the same key are received within
 // the interval, the timer is reset.
-type Batcher struct {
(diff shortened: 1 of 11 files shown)


ItalyPaleAle (commented):
LGTM but please note this is a breaking change for all packages that depend on events/queue

yaron2 (approved):

PR #85, opened 2024-02-26
Adds events/batch/singular

JoshVanL:
Adds Singular Batcher which treat all events as the same event key. Subscribers provide a function when subscribing which will be executed once the batch triggers.

Used by https://github.com/dapr/components-contrib/pull/3363

### DIFF
--- a/events/batcher/singuler.go
+++ b/events/batcher/singuler.go
@@ -21,7 +21,6 @@ import (
 
 // Singular is a Batcher which batches events and treats them as all the same
 // key, behaving like a singular batched queue.
-// Subscribe
 type Singular struct {
 	b       *Batcher[int]
 	closeCh chan struct{}


events/batcher/singuler.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// Singular is a Batcher which batches events and treats them as all the same
+// key, behaving like a singular batched queue.
+// Subscribe
```
ItalyPaleAle, line 24: Typo?
```suggestion
```

ItalyPaleAle (dismissed):
I don't love the name Singular, but this is a bit of a nit.

Approved, feel free to merge if you don't have further changes to make

ItalyPaleAle (approved):

PR #87, opened 2024-03-04
Updates go version to 1.22

JoshVanL:
Chore


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/dapr/kit
 
-go 1.20
+go 1.22
```
ItalyPaleAle, line 3: There's no need to update the Go version in dapr/kit for the sake of updating it, unless we need to depend on features added in 1.21+?

ItalyPaleAle (changes requested):

cicoyle:
I think this can be closed due to this being merged: https://github.com/dapr/kit/pull/105

PR #88, opened 2024-03-06
feat: enable original key to be returned for metadata property fields

sicoyle:
# Description

components contrib metadata updates needs us to return the original key back here

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/metadata/utils.go
+++ b/metadata/utils.go
@@ -23,18 +23,17 @@ import (
 )
 
 // GetMetadataProperty returns a property from the metadata map, with support for case-insensitive keys and aliases.
-func GetMetadataProperty(props map[string]string, keys ...string) (val string, ok bool) {
+func GetMetadataProperty(props map[string]string, keys ...string) (key string, val string, ok bool) {
 	lcProps := make(map[string]string, len(props))
 	for k, v := range props {
 		lcProps[strings.ToLower(k)] = v
 	}
 	for _, k := range keys {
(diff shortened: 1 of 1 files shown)


metadata/utils.go
```diff
@@ -23,18 +23,17 @@ import (
 )
 
 // GetMetadataProperty returns a property from the metadata map, with support for case-insensitive keys and aliases.
-func GetMetadataProperty(props map[string]string, keys ...string) (val string, ok bool) {
+func GetMetadataProperty(props map[string]string, keys ...string) (key string, val string, ok bool) {
```
ItalyPaleAle, line 26: Please create a separate method `GetMetadataPropertyWithMatchedKey` (or something else) rather than modifying the exiting API. This way, we don't have to change the API in dozens of places!

ItalyPaleAle (changes requested):

metadata/utils.go
ItalyPaleAle: Once `GetMetadataPropertyWithMatchedKey` is implemented, `GetMetadataProperty` can change to:

```go
func GetMetadataProperty(props map[string]string, keys ...string) (val string, ok bool) {
_, val, ok = GetMetadataPropertyWithMatchedKey(props, keys...)
return val, ok
}

sicoyle:
ready

ItalyPaleAle (approved):

PR #99, opened 2024-07-22
concurrency/mutexmap: Adds DeleteUnlock

JoshVanL:


yaron2 (approved):

concurrency/mutexmap.go
```diff
@@ -106,6 +108,16 @@ func (a *mutexMap[T]) Delete(key T) {
 	a.lock.Unlock()
 }
 
+func (a *mutexMap[T]) DeleteUnlock(key T) {
+	a.lock.Lock()
+	mutex, ok := a.items[key]
+	delete(a.items, key)
```
yaron2, line 114: Shouldn't the item be deleted only if ok is true? Otherwise two calls to this function might see the first deleting, releasing the lock and then the second attemping to delete an already deleted item.

ItalyPaleAle: I think this should be ok, but it may not hurt to do as you suggest anyways. Calling `delete` with a key that doesn't exist is a no-op.

yaron2: If it's a no-op then we should be good, thanks for confirming

artursouza (approved):

PR #101, opened 2024-08-19
concurrency/mutexmap Move Unlock to after operation

JoshVanL:
# Description

_Please explain the changes you've made_

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/concurrency/mutexmap.go
+++ b/concurrency/mutexmap.go
@@ -58,50 +58,57 @@ func (a *mutexMap[T]) Lock(key T) {
 	a.lock.RLock()
 	mutex, ok := a.items[key]
 	a.lock.RUnlock()
-	if !ok {
-		a.lock.Lock()
-		mutex, ok = a.items[key]
-		if !ok {
-			mutex = &sync.RWMutex{}
-			a.items[key] = mutex
-		}
-		a.lock.Unlock()
+	if ok {
+		mutex.Lock()
+		return
 	}
+
+	a.lock.Lock()
+	mutex, ok = a.items[key]
+	if !ok {
+		mutex = &sync.RWMutex{}
+		a.items[key] = mutex
+	}
+	a.lock.Unlock()
 	mutex.Lock()
 }
 
 func (a *mutexMap[T]) Unlock(key T) {
 	a.lock.RLock()
 	mutex, ok := a.items[key]
-	a.lock.RUnlock()
 	if ok {
 		mutex.Unlock()
 	}
+	a.lock.RUnlock()
 }
 
 func (a *mutexMap[T]) RLock(key T) {
 	a.lock.RLock()
 	mutex, ok := a.items[key]
 	a.lock.RUnlock()
-	if !ok {
-		a.lock.Lock()
-		mutex, ok = a.items[key]
-		if !ok {
-			mutex = &sync.RWMutex{}
-			a.items[key] = mutex
-		}
-		a.lock.Unlock()
+
+	if ok {
+		mutex.RLock()
+		return
 	}
+
+	a.lock.Lock()
+	mutex, ok = a.items[key]
+	if !ok {
+		mutex = &sync.RWMutex{}
+		a.items[key] = mutex
+	}
+	a.lock.Unlock()
 	mutex.RLock()
 }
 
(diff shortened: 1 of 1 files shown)


concurrency/mutexmap.go
```diff
@@ -58,50 +58,57 @@ func (a *mutexMap[T]) Lock(key T) {
 	a.lock.RLock()
 	mutex, ok := a.items[key]
 	a.lock.RUnlock()
+	if ok {
```
artursouza, line 61: Is this just a style change? It seems to do the same as before.

concurrency/mutexmap.go
```diff
@@ -58,50 +58,57 @@ func (a *mutexMap[T]) Lock(key T) {
...
 }
 
 func (a *mutexMap[T]) Unlock(key T) {
 	a.lock.RLock()
 	mutex, ok := a.items[key]
-	a.lock.RUnlock()
 	if ok {
 		mutex.Unlock()
 	}
+	a.lock.RUnlock()
```
artursouza, line 82: Why is this needed? The lock of the map is now bundled with the lock of the mutex itself. The original design was keeping those separate.

artursouza (changes requested):
Can you provide more context of why this change is needed? The previous design separated the lock of the map with the lock of each mutex.

concurrency/mutexmap.go
JoshVanL: The change aims to release the outer lock after the inner operation has completed. This prevents race conditions on inner operations of the same key

concurrency/mutexmap.go
artursouza: OK. I am fine with either way.

concurrency/mutexmap.go
```diff
@@ -58,50 +58,57 @@ func (a *mutexMap[T]) Lock(key T) {
 	a.lock.RLock()
 	mutex, ok := a.items[key]
 	a.lock.RUnlock()
+	if ok {
+		mutex.Lock()
```
artursouza, line 62: Don't you need to also move this before the map's unlock?

artursouza (changes requested):

concurrency/mutexmap.go
JoshVanL: No, we have to release the outer lock before locking the inner lock else we will get a deadlock. If the mutex is already locked, other operations like `DeleteUnlock` would never complete.

artursouza:
Can we have a unit test that runs multiple operations in parallel to try a deadlock scenario? I don't disagree that there is a bug but I am concerned about this PR as well for bundling the 2 locks instead of keeping them separate.

concurrency/mutexmap.go
artursouza: I find it odd that we need to bundle the unlock of item with the unlock of the map.

artursouza: Oh. I think I get it now. It is because of the delete.

concurrency/mutexmap.go
elena-kolevska: I'm not so worried about `DeleteUnlock` because it will never be called before the initial Lock has completed and returned. I think the bigger problem is when we have a delete and a concurrent lock again. Then we would actually get two different mutexes controlling the same resource.

JoshVanL:
Because it is a true race condition of LOC and execution, I'm not sure of a way to write a sensible unit test for it.

artursouza:
It is hard to make a deterministic test for this. Is it possible to make a test that will not cause false positives? Meaning that the race condition will most likely cause the test to fail but not 100% guaranteed. On the other hand, not having the race condition will make the test pass 100%. This way, we can run the test a few times to make sure. It is not ideal but better than visual inspection (aka code review) IMO.

Is there another layer (runtime, maybe) where it can be tested?

JoshVanL:
I can look into doing that. This bug is currently being manifested in Dapr integration tests failing.

yaron2 (approved):

artursouza (approved):
Approved under the assumption that integration tests will show the fix.

artursouza (approved):

PR #108, opened 2024-11-12
Adds generic ring

JoshVanL:
Adds a generic implementation of the stdblib ring buffer so that each ring `Value` can be a concrete type.
https://pkg.go.dev/container/ring

Adds a `ring/Buffered` which implements a ring which buffers nodes, and expands/contracts based on live nodes. Used to reduce memory allocations.

Adds `Len() int` and `Keys() []K` func to the generic Map cmap.

Changes `events/queue` Processor `Queueable` to be an exported type. No functional change, but consumed types should be exported.

### DIFF
--- a/ring/ring.go
+++ b/ring/ring.go
@@ -1,0 +1,137 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package ring implements operations on circular lists.
+// Edited to be generic.
+package ring
+
+// A Ring is an element of a circular list, or ring.
+// Rings do not have a beginning or end; a pointer to any ring element
+// serves as reference to the entire ring. Empty rings are represented
+// as nil Ring pointers. The zero value for a Ring is a one-element
+// ring with a nil Value.
+type Ring[T any] struct {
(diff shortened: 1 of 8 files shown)


ring/ring.go
```diff
@@ -0,0 +1,137 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
```
elena-kolevska, line 3: Do we need to add a mention about Dapr editing the file? I don't know what the convention is here.

elena-kolevska (approved):

ring/ring.go
JoshVanL: That's done 3 lines down. I assumed the editor being Dapr was assumed through context.

elena-kolevska (approved):
Really cool stuff!! Maybe we can also consider a dynamic buffer size that starts from a low value and grows exponentially, up to a max value.

ring/ring.go
elena-kolevska: Yeah, I saw, was just wondering if the name Dapr needs to be mentioned, I genuinely don't know if that's the case.

yaron2 (approved):

PR #109, opened 2024-11-22
Add error code and category funcs

jake-engelberg:
# Description

In order to support error code metrics, sometimes the category will need to be passed along with a kit error, as well as the ability to retrieve it and its error code.

## Issue reference
Please reference the issue this PR will close: https://github.com/dapr/proposals/pull/67

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -56,6 +56,9 @@ type Error struct {
 
 	// Tag is a string identifying the error, used with HTTP responses only.
 	tag string
+
+	// Category is a string identifying the category of the error, used for error code metrics only.
+	category string
 }
 
 // ErrorBuilder is used to build the error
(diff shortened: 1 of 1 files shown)


nelson-parente (approved):

elena-kolevska (changes requested):
Can you please add a unit test for `ErrorCode` ?
Also, if I remember correctly, we should be checking if the `tag` field has a value first, before checking `ErrorInfo`.

errors/errors.go
```diff
@@ -56,6 +56,9 @@ type Error struct {
 
 	// Tag is a string identifying the error, used with HTTP responses only.
 	tag string
+
+	// Category is a string identifying the category of the error, used for error code metrics only.
```
cicoyle, line 60:
```suggestion
	// Category is a string identifying the category of the error, for example: ___, ___, ____, used for error code metrics only. 
```
small nit: can you expand here on types of categories to be clearer. Then, looks like you'll need to tweak this file too `errors/codes.go`. Otherwise, the changes look fine :)

elena-kolevska (approved):
Great work! 
We ended up doing the check for the legacy tag twice (you had it right the first time), but that's ok, I don't want to delay this PR.

PR #110, opened 2025-01-06
Adds events/broadcaster

JoshVanL:
Adds a generic buffered dmessage broadcaster which will relay a typed message to a dynamic set of subscribers.

### DIFF
--- a/events/broadcaster/broadcaster.go
+++ b/events/broadcaster/broadcaster.go
@@ -20,6 +20,8 @@ import (
 	"sync/atomic"
 )
 
+const bufferSize = 10
+
 type eventCh[T any] struct {
 	id           uint64
 	ch           chan<- T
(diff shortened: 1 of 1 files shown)


events/broadcaster/broadcaster.go
```diff
@@ -0,0 +1,132 @@
...
+			select {
+			case <-ctx.Done():
+				return
+			case <-b.closeCh:
+				return
+			case env := <-bufferedCh:
+				select {
+				case ch <- env:
+				case <-ctx.Done():
+				case <-b.closeCh:
```
filintod, line 99: what is the reason to not go direct `ch <- bufferedCh`? if `ch` is blocked (not been read), wouldn't it be in the same place if this is inside the embedded select as being on the outer one?

events/broadcaster/broadcaster.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+}
+
+// Close closes the Broadcaster. It blocks until all events have been sent to
+// the subscribers. The Broadcaster will be a no-op after this call.
+func (b *Broadcaster[T]) Close() {
+	defer b.wg.Wait()
+	b.lock.Lock()
+	if b.closed.CompareAndSwap(false, true) {
+		fmt.Printf(">>HERE!!!!\n")
```
filintod, line 128: extra debug stmt

JoshVanL: Thank

events/broadcaster/broadcaster.go
JoshVanL: Thanks, done

yaron2 (approved):

PR #115, opened 2025-02-24
concurrency/lock: Adds context and outercancel locks

JoshVanL:
Adds context lock which will cancel and return an error to Lock & RLock if the given context cancels before the lock is achieved.

Adds outercancel lock which will cancel all RLocks in progress if the outer Lock is called.

### DIFF
--- a/concurrency/lock/outercancel.go
+++ b/concurrency/lock/outercancel.go
@@ -37,8 +37,9 @@ type holdresp struct {
 }
 
 type OuterCancel struct {
-	ch        chan *hold
-	cancelErr error
+	ch              chan *hold
+	cancelErr       error
+	gracefulTimeout time.Duration
 
 	lock chan struct{}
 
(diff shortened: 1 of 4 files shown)


concurrency/lock/outercancel.go
```diff
@@ -0,0 +1,197 @@
...
+		}
+	} else {
+		o.lock <- struct{}{}
+	}
+
+	o.rcancelLock.Lock()
+
+	if h.writeLock {
+		for _, cancel := range o.rcancels {
+			go cancel()
```
yaron2, line 108: Could there be a case where if any of these goroutines hangs we'll have a deadlock?

JoshVanL: Yes- it's up to the consumer to ensure that they don't introduce a deadlock by making the users of the RLock correctly respect the context that they are given. As is the same is true for consuming a normal `sync.RWMutex`.

yaron2 (approved):

PR #119, opened 2025-04-22
events/loop & slices

JoshVanL:
Adds a generic control loop implementation to `event/loop`.

Adds a new `slices` package that provides a generic slice de-duplication func.

Makes events batcher and queue processer taker in Options.

Allows enqueuing multiple processor items in same func call.

### DIFF
--- a/events/queue/processor.go
+++ b/events/queue/processor.go
@@ -18,8 +18,14 @@ import (
 	"sync/atomic"
 	"time"
 
+	"k8s.io/utils/clock"
 	kclock "k8s.io/utils/clock"
 )
+
+type Options[K comparable, T Queueable[K]] struct {
+	ExecuteFn func(r T)
+	Clock     kclock.Clock
+}
 
 // Processor manages the queue of items and processes them at the correct time.
 type Processor[K comparable, T Queueable[K]] struct {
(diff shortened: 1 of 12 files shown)


yaron2 (approved):

yaron2 (approved):

slices/slices.go
```diff
@@ -0,0 +1,27 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package slices
+
+// Deduplicate removes duplicate elements from a slice.
+func Deduplicate[S ~[]E, E comparable](s S) S {
```
antontroshin, line 17: Is this essentially the [slices.Compact](https://pkg.go.dev/slices#Compact)? Can use it instead?

JoshVanL: @antontroshin Compat only does consecutive elements whereas this does the full slice (via map indexing).

events/queue/processor.go
```diff
@@ -36,40 +41,45 @@ type Processor[K comparable, T Queueable[K]] struct {
...
-// WithClock sets the clock used by the processor. Used for testing.
-func (p *Processor[K, T]) WithClock(clock kclock.Clock) *Processor[K, T] {
-	p.clock = clock
-	return p
-}
-
 // Enqueue adds a new item to the queue.
 // If a item with the same ID already exists, it'll be replaced.
-func (p *Processor[K, T]) Enqueue(r T) {
+func (p *Processor[K, T]) Enqueue(rs ...T) {
```
nelson-parente, line 61: nit: Enqueue adds new items to the queue.

nelson-parente (approved):

cicoyle (approved):

PR #120, opened 2025-05-11
events/loop: add reset

JoshVanL:
Update loop implementation is include functionality for Reset which is useful when caching the loop struct for future use to reduce allocations.


cicoyle (approved):
lgtm

JoshVanL:
Thanks @cicoyle

PR #130, opened 2025-07-08
Update deps

javier-aliaga:
# Description

Update go to version 1.24.3
Update golangci-lint to v1.64.8
Fixes all lint issues

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


concurrency/lock/outercancel_test.go
```diff
@@ -162,7 +162,7 @@ func Test_OuterCancel(t *testing.T) {
 		t.Parallel()
 
 		l := NewOuterCancel(terr, time.Second)
-
+		//nolint:usetesting
```
acroca, line 165: Why can't this one use `t.Context()`?

context/pool.go
```diff
@@ -52,7 +52,10 @@ func NewPool(ctx ...context.Context) *Pool {
 	go func() {
 		defer cancel()
 		defer p.lock.RUnlock()
-		for i := 0; i < len(p.pool); i++ {
+		for i := range len(p.pool) {
+			if p.pool == nil {
```
acroca, line 56: `p.pool` is not going to be nil because it's defined in line 39. Is this for the linter? If necessary, can this go outside the loop?

cron/doc.go
```diff
@@ -36,7 +35,9 @@ them in their own goroutines.
 # Time mocking
 
 import (
+
 	clocktesting "k8s.io/utils/clock/testing"
+
```
acroca, line 40: Is this expected?

context/pool.go
javier-aliaga: it can be null, when you close the pool it is assigned to nil https://github.com/dapr/kit/blob/34f8820d2a4a8e4de66cd0abe2df008dec458b1d/context/pool.go#L89

So then this loop will panic

cron/doc.go
javier-aliaga: checked, lint --fix added it

